Issue #15, opened 2026-03-02
bug: edit_core_memory appends instead of replacing


prakashUXtech:
## Problem

\`Soul.edit_core_memory(field, value)\` appends the new value to existing content instead of replacing it. After multiple edits, the core memory accumulates stale data.

```python
await soul.edit_core_memory("persona", "I am a coding assistant")
await soul.edit_core_memory("persona", "I am a Python expert")
# persona is now "I am a coding assistant I am a Python expert" 
# instead of just "I am a Python expert"
```

## Expected

\`edit_core_memory\` should replace the field value, not append. If append behavior is desired, it should be an explicit \`append_core_memory\` method.

## Impact

LOW — Core memory editing works but accumulates cruft. Workaround: manually clear before editing.

**Labels**: bug, priority-low, memory

PR #33, opened 2026-03-06
feat: memory evolution - bug fix, archival tier, compression, temporal graph

prakashUXtech:
## Summary

- **Bug #15 fix**: `edit_core_memory` now replaces values instead of appending. The old append behavior lives on as a new `append()` method for explicit incremental updates.
- **Archival memory tier**: `ArchivalMemoryStore` stores compressed conversation archives with keyword search across summaries/key moments and date-range queries.
- **Compression pipeline**: `MemoryCompressor` provides rule-based (no LLM) summarization, token-overlap deduplication, importance/age pruning, and inline vs external splitting for `.soul` export.
- **Temporal knowledge graph**: Edges now carry `valid_from`/`valid_to` timestamps. New methods: `as_of_date()` for point-in-time queries, `relationship_evolution()` for full history, `expire_relationship()` to end a relationship. Backward-compatible with old edge format.
- **Simulation script**: `scripts/simulate_memory_evolution.py` runs 100 interactions and demonstrates all three systems working together with printed stats.

## Test plan

- [x] 4 regression tests for Bug #15 (`tests/test_bug_15_core_memory.py`)
- [x] 14 tests for archival tier (`tests/test_archival.py`)
- [x] 16 tests for compression pipeline (`tests/test_compression.py`)
- [x] 20 tests for temporal graph (`tests/test_temporal_graph.py`)
- [x] Updated existing `test_core_memory_edit` to match new replace behavior
- [x] Full suite: 124 tests passing, 0 failures
- [x] Simulation script runs end-to-end successfully

### DIFF
--- a/scripts/simulate_memory_evolution.py
+++ b/scripts/simulate_memory_evolution.py
@@ -1,0 +1,233 @@
+#!/usr/bin/env python3
+# simulate_memory_evolution.py — Simulates 100 interactions to demonstrate
+# memory compression, archival storage, and temporal graph evolution.
+# Created: 2026-03-06
+
+"""
+Simulate 100 interactions and show:
+  - Memory compression working (before/after counts)
+  - Archival storage growing
+  - Temporal graph evolution
+  - Final stats summary
+"""
+
+from __future__ import annotations
+
+import asyncio
+import uuid
+from datetime import datetime, timedelta
+
+from soul_protocol.memory.archival import ArchivalMemoryStore, ConversationArchive
+from soul_protocol.memory.compression import MemoryCompressor
+from soul_protocol.memory.graph import KnowledgeGraph
+from soul_protocol.soul import Soul
+from soul_protocol.types import Interaction, MemoryEntry, MemoryType
+
+
+# Simulated conversation topics for variety
+TOPICS = [
+    ("python", "I use Python for data analysis", "technology"),
+    ("rust", "I'm learning Rust for systems programming", "technology"),
+    ("cooking", "I love cooking Italian food", "hobby"),
+    ("music", "I play guitar in a band", "hobby"),
+    ("work", "I work at a startup called Acme", "career"),
+    ("travel", "I live in Tokyo and love it", "location"),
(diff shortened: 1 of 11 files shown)


prakashUXtech:
## Self-Review: PR #33 — Memory Evolution

### What's Good
- Bug #15 fix is correct — `edit()` now replaces, old append behavior preserved as `append()`
- Regression test written first (TDD), 4 tests covering the fix
- Archival memory tier (ConversationArchive) is clean — keyword search across summaries + key_moments
- Compression pipeline is rule-based (no LLM needed) — dedup, pruning, export splitting all work
- Temporal graph adds `valid_from`/`valid_to` to edges with `as_of_date()` and `relationship_evolution()`
- Simulation script validates everything end-to-end
- Core/engine boundary respected — zero core/ imports, core/ untouched

### Issues Found

1. **`manager.py` docstring still says "Append to core memory fields"** — Should be updated to "Replace core memory fields" after the bug fix. Misleading documentation.

2. **`compression.py` imports `from soul_protocol.types import MemoryEntry`** — Uses absolute import. Should be `from ..types import MemoryEntry` for consistency with the rest of the memory module.

3. **`TemporalEdge.valid_to` is `datetime | None`** — When `None`, the relationship is considered "current". This is correct but should be documented in the class docstring to avoid confusion.

4. **`search_archives` token overlap threshold is implicit** — No similarity threshold parameter; returns any non-zero overlap. Consider adding a `min_score` parameter like the memory search has.

5. **`prune_by_importance` uses `datetime.now()` for age calculation** — Should use UTC-aware datetime to match other timestamp handling. Also, if `MemoryEntry.timestamp` is naive but `datetime.now()` returns a different timezone, the comparison could fail.

6. **Branch appears to diff against a very old base** — The `--name-only` shows nearly every file in the repo. Verify this branch is based on current `main` HEAD, not an ancient commit.

### Verdict
**Approve with minor fixes needed.** The bug fix is solid, the new features are well-tested (57 tests), and the engine boundary is clean. Fix the docstring and verify the branch base before merge.

PR #91, opened 2026-03-13
improve(dedup): add synonym expansion to Jaccard similarity

goyalpalak18:
## Description

I noticed an issue where the deduplication pipeline was generating false negatives because it didn't apply synonym expansion, unlike our search pipeline. Specifically, `_jaccard_similarity()` was relying on raw `tokenize()` outputs. This meant semantically identical facts that used different vocabulary (e.g., "bug" vs "error", "login" vs "auth") were returning low Jaccard scores and being stored as separate, new facts. Left unchecked, this causes memory to bloat with near-duplicates over time.

To fix this, I updated the tokenization step in `_jaccard_similarity()` to use the existing `_expand_synonyms()` function from `search.py`. 

Before this change, comparing a fact like "User reported a bug in the login flow" against "User reported an error in the authentication flow" gave a Jaccard score of 0.50, triggering a false `CREATE`. Now, the expanded token sets correctly evaluate to a score of 1.0, triggering a `SKIP`. The existing thresholds (0.6 and 0.85) still work perfectly with this expanded token set without causing false positives for genuinely unrelated facts.

## Changes

* **Modified `src/soul_protocol/runtime/memory/dedup.py`**: Imported `_expand_synonyms` and wrapped the `tokenize(a)` and `tokenize(b)` calls inside `_jaccard_similarity()`.
* **Added `tests/test_memory/test_dedup.py`**: Created a new test file containing 16 test cases to verify synonym-aware similarity and `reconcile_fact` decisions.

## Testing

I added comprehensive tests in the new file that specifically cover:
* Synonym-variant facts correctly triggering SKIP or MERGE.
* Unrelated facts correctly triggering CREATE despite the synonym expansion.
* Edge cases involving empty strings, single-token facts, and facts lacking matchable synonyms.
* Facts that share synonyms but remain genuinely distinct, ensuring no false positives are introduced.

### DIFF
--- a/src/soul_protocol/runtime/memory/dedup.py
+++ b/src/soul_protocol/runtime/memory/dedup.py
@@ -4,10 +4,13 @@
 #   - SKIP: >0.85 similarity (near-duplicate), MERGE: 0.6-0.85 (update existing),
 #     CREATE: <0.6 (genuinely new fact)
 #   - Uses the same tokenize() function as search.py for consistency
+# Updated: 2026-03-20 — Asymmetric synonym expansion in _jaccard_similarity():
+#   only the existing fact (b) is expanded, mirroring relevance_score() in search.py.
+#   Prevents false positives when short facts share synonym-group vocabulary.
 
 from __future__ import annotations
 
-from soul_protocol.runtime.memory.search import _expand_synonyms, tokenize
+from soul_protocol.runtime.memory.search import expand_synonyms, tokenize
 from soul_protocol.runtime.types import MemoryEntry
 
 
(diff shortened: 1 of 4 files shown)


goyalpalak18:
Hi @prakashUXtech ! Could you please have a look at this when you have a moment? Thanks!

prakashUXtech (approved):
Looks good, clean fix for a real problem. The before/after on the bug-vs-error example (0.33 → 1.0) makes the case clearly.

One small thing to keep in mind: `_expand_synonyms` is a private function that's now imported in three places (search.py internally, strategy.py, and now dedup.py). It works fine and matches the existing pattern in strategy.py, but if that function ever gets renamed or refactored, three consumers break quietly. Might be worth dropping the underscore prefix at some point to signal it's part of the internal API contract. Not blocking, just something to watch.

Tests look thorough — the "no false positives for unrelated facts" cases are the ones that matter most and they're well covered.

PR #161, opened 2026-04-13
feat(spec): RetrievalTrace receipt for every recall

prakashUXtech:
New DSP primitive that sits alongside `MemoryEntry` and `Interaction`. `RetrievalTrace` is the receipt for one lookup — query, candidates with scores, who asked, how long it took. Scope-free by design; downstream systems (paw-runtime log sink, graduation policy, compliance reporter, SoulBench fixture generator) all read the same shape.

First PR of **Move 4 — Retrieval Trace + Graduation**. Three follow-ups: paw-runtime log sink + CLI (PR-B), graduation policy (PR-C), Why? drawer surface (PR-D).

### \`spec/retrieval.py\` (new)
- \`RetrievalTrace\` — the portable receipt: \`query\`, \`actor\`, \`source\`, \`candidates\`, \`picked\`, \`used_by\`, \`latency_ms\`, \`pocket_id\`, \`timestamp\`, free-form \`metadata\`.
- \`RetrievalCandidate\` — per-result \`id\` + \`source\` + \`score\` + optional \`tier\`. \`score\` is runtime-defined; scale is not normalised across sources (callers that compare cross-source should normalise).
- \`mark_used(picked_ids, used_by)\` — separate from construction so retrieval functions return a trace with candidates populated, and downstream callers fill in the acted-on subset without reconstructing the receipt.
- Exported from \`soul_protocol.spec\`.

### \`runtime/soul.py\`
- \`Soul.recall()\` and \`Soul.smart_recall()\` build a \`RetrievalTrace\` on every call (including empty-results) and expose it via the read-only \`Soul.last_retrieval\` property. In-memory only — explicitly dropped on export/awaken. No public API break for existing callers that don't read \`last_retrieval\`.
- \`_resolve_actor(soul)\` — defensive helper so test doubles that stub out \`_identity\` still get a trace with a non-empty shape.
- \`_build_trace()\` — tolerates both the spec \`MemoryEntry\` (no \`importance\`, no \`type\`) and the runtime \`MemoryEntry\` (both present). Falls back to default score (0.5) when importance is absent.
- \`smart_recall\` adds variant metadata (\`{variant: "smart_recall", reranked: bool, candidate_pool: N}\`) so downstream consumers can distinguish reranked vs heuristic traces without parsing the query string.

## Why this primitive, now

Three downstream problems share the same missing piece — a receipt for every lookup:
1. **Debug** — "why did the agent surface X?" is unanswerable today. With receipts it's readable.
2. **Graduation** — the bespoke 3×-same-path rule in \`correction_soul_bridge\` (pocketpaw #929) collapses to a general policy once access counts are logged.
3. **Compliance + eval** — "show me every memory actor X saw last quarter" becomes a single filter over the log. SoulBench fixtures replay real traces.

Landing the primitive first means the log-sink PR and the graduation PR both read the same shape from day one.

## Test plan

- [x] 12 new tests in \`tests/test_retrieval_trace.py\` covering model defaults + round-trip + mark_used contract + recall/smart_recall emission + requester_id actor precedence + empty-results path + export/awaken drops the in-memory field + spec-shaped memories flow through \`_build_trace\` without raising
- [x] \`pytest tests/\` — 2117 passed (was 2105), 0 failed
- [x] \`ruff check\` — clean

## Context

Design doc: paw-enterprise \`docs/enterprise/RETRIEVAL-TRACE.md\` (paw-enterprise PR #67 for the doc itself).

### DIFF
--- a/src/soul_protocol/runtime/soul.py
+++ b/src/soul_protocol/runtime/soul.py
@@ -1,4 +1,7 @@
 # soul.py — The main Soul class: birth, awaken, observe, save, export
+# Updated: 2026-04-14 (v0.3.1) — remember() accepts a ``scope`` kwarg so callers
+#   (notably SoulFactory.from_template) can stamp RBAC/ABAC tags on seeded
+#   memories. Pairs with spec/scope.match_scope on the recall side.
 # Updated: feat/mcp-sampling-engine — Added set_engine() to swap CognitiveEngine at runtime.
 #   MCPSamplingEngine uses this for lazy wiring on first MCP tool call.
 # Updated: 2026-03-22 — Added learn() and learning_events for LearningEvent pipeline.
@@ -61,6 +64,7 @@ from pathlib import Path
 from typing import Any
 
 from soul_protocol.spec.learning import LearningEvent
+from soul_protocol.spec.trace import RetrievalTrace, TraceCandidate
 
 from .bond import Bond
 from .cognitive.engine import CognitiveEngine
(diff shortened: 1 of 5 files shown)


prakashUXtech (commented):
## Review — feat(spec): RetrievalTrace receipt for every recall

**Verdict: LGTM with one non-blocking note.** This is a well-scoped primitive. The shape is minimal, the score approximation is honest about its limitations, the in-memory-only contract is clearly documented, and the 12 tests cover the right surface area including the empty-results path and the export/awaken drop guarantee.

---

### What's good

The separation between construction and `mark_used()` is the right call. Retrieval functions emit a trace with candidates populated; downstream callers fill in `picked` and `used_by` when they know what was actually consumed. That keeps the receipt portable across log-sink, graduation policy, and SoulBench without coupling the receipt shape to any one consumer.

The `_build_trace` helper's tolerance of both spec and runtime `MemoryEntry` shapes (via `getattr` with safe defaults) is exactly the kind of defensive glue that prevents prod surprises when the two shapes diverge.

Documenting the score approximation (`importance / 10`) inline in `_build_trace`, with an explicit call-out that paw-runtime's log sink is free to override it, is the right way to ship an imperfect heuristic without misleading consumers.

---

### Findings (1 — non-blocking)

**`timestamp` uses `datetime.now()` (local time) instead of `datetime.utcnow()` or `datetime.now(UTC)`.**

`spec/retrieval.py`, `RetrievalTrace.timestamp` field:

```python
timestamp: datetime = Field(default_factory=datetime.now)
```

`datetime.now()` produces a timezone-naive local datetime. The paw-runtime JSONL sink and any compliance reporter that joins traces across machines or timezones will get inconsistent orderings. The spec `MemoryEntry` has the same pattern (`ingested_at`), but `RetrievalTrace` is explicitly designed for cross-system log aggregation — the pain will be felt sooner here.

Suggested fix for the field default:

```python
from datetime import UTC
timestamp: datetime = Field(default_factory=lambda: datetime.now(UTC))
```

If changing the default mid-stream is too noisy, at minimum add a note in the class docstring that consumers should treat the timestamp as local time and normalize before cross-host comparison. Worth tracking this as a follow-up if not fixed now.

---

### Test gaps

Coverage is solid. One missing case worth adding before PR-B lands: `mark_used()` called with IDs that are not in `candidates` should not raise — the spec docstring implies the caller fills in whatever subset they used, but there's no test asserting the model doesn't validate membership. If the graduation policy or compliance reporter generates a trace with a `picked` ID that wasn't in the original candidate set (e.g., after a rerank reorders IDs), a silent pass is safer than a surprise validation error.

---

### Nit

The `_resolve_actor` module-level helper is only called by `Soul` methods. A leading underscore is correct, but consider whether it belongs closer to the methods that use it (as an inner function or a staticmethod on `Soul`) rather than at module level. Not a blocker — current placement keeps `_build_trace` testable independently, which the test suite takes advantage of.

---

**Targets `main` — worth confirming this should be against `dev` before merge.**

prakashUXtech (commented):
test

PR #162, opened 2026-04-13
feat(spec): scope tags on MemoryEntry + match_scope helper

prakashUXtech:
First foundational primitive for the RBAC/ABAC layer described in the fleet vision. \`MemoryEntry\` gains an optional \`scope: list[str]\` field; \`Soul.recall\` accepts an optional \`scopes=\` kwarg; a standalone \`match_scope\` helper centralises hierarchical glob matching so every retrieval path uses the same logic once paw-runtime adopts it in PR-B.

Move 5 PR-A. Three follow-ups: PR-B (Fabric + Pocket scope + policy engine in pocketpaw), PR-C (scope chip UI), PR-D (E2E enforcement test).

### \`spec/memory.py\`
- New \`scope: list[str]\` field on \`MemoryEntry\`. Default empty list = "no scope assigned, visible to anyone." Round-trips through \`model_dump\` cleanly.

### \`spec/scope.py\` (new)
- \`match_scope(entity_scopes, allowed_scopes)\` — boolean OR over cartesian; empty either side passes through (back-compat for pre-scope memories AND callers without a scope filter). Hierarchical glob: \`org:sales:*\` matches \`org:sales:leads\` and \`org:sales:deals\`, \`*\` matches everything.
- \`normalise_scopes(scopes)\` — strip + lowercase + dedupe, drops empty + non-string entries.

### \`runtime/types.py\`
- Same \`scope\` field on the runtime \`MemoryEntry\` so producers can attach scope at remember-time once PR-B lands. Spec and runtime entries diverge on tier-related fields but share the scope shape exactly.

### \`runtime/soul.py\`
- \`Soul.recall\` accepts \`scopes: list[str] | None\`. Empty/None preserves existing behaviour. When set, results are filtered after scoring — underlying retrieval order is unchanged. Filter pulls scope via \`getattr\` so spec-only test doubles keep working.

## Test plan

- [x] 18 new tests in \`tests/test_scope_tags.py\` covering: exact match, no-overlap denial, hierarchical glob, segment-boundary rejection, star wildcard, both-empty back-compat, multiple-scope OR, unknown-pattern denial, normalise round-trip, default empty list, serialization round-trip, recall back-compat, recall+scopes filtering unscoped vs scoped, runtime entry scope round-trip.
- [x] \`pytest tests/\` — 2123 passed (was 2117), 0 failed
- [x] \`ruff check\` — clean

## Design contract

The match semantics are designed to be **fail-open for back-compat, fail-closed once scope is set**:
- Memory with no scope → visible to any caller (pre-scope memories aren't broken).
- Caller with no scope filter → sees everything (unchanged retrieval).
- Memory WITH scope + caller WITH scope → only matching entries surface.

This means rolling out scopes is incremental: tag the memories that need protection, leave the rest unscoped, and only callers that explicitly pass \`scopes=\` get filtering.

## Context

Design referenced in the fleet platform brainstorm — \`scope\` is one of the three lean primitives (along with RetrievalTrace from Move 4 and Soul Templates in Move 6) that compose into the Sarah-the-VP-Sales scenario.

### DIFF
--- a/src/soul_protocol/runtime/soul.py
+++ b/src/soul_protocol/runtime/soul.py
@@ -763,9 +763,22 @@ class Soul:
         requester_id: str | None = None,
         bond_strength: float | None = None,
         bond_threshold: float = 30.0,
+        progressive: bool = False,
+        scopes: list[str] | None = None,
     ) -> list[MemoryEntry]:
-        """Soul recalls relevant memories with visibility filtering."""
-        effective_bond = bond_strength if bond_strength is not None else self._identity.bond.bond_strength
+        """Soul recalls relevant memories with visibility + scope filtering.
+
+        ``scopes`` applies hierarchical RBAC/ABAC matching. When omitted (or
+        empty) the caller sees everything they would otherwise see — back-compat
+        for pre-scope memories and consumers that do not need RBAC. When
+        present, only memories whose own ``scope`` list intersects with the
+        caller's ``scopes`` (via :func:`soul_protocol.spec.match_scope`) are
+        returned. Filtering happens after scoring so the underlying recall
+        order is unchanged.
+        """
+        effective_bond = (
+            bond_strength if bond_strength is not None else self._identity.bond.bond_strength
+        )
         results = await self._memory.recall(
             query=query,
             limit=limit,
@@ -775,6 +788,13 @@ class Soul:
             bond_strength=effective_bond,
             bond_threshold=bond_threshold,
         )
+        if scopes:
+            from soul_protocol.spec.scope import match_scope
+
+            results = [
+                entry for entry in results
+                if match_scope(getattr(entry, "scope", None), scopes)
+            ]
         if not results:
             logger.debug("Recall returned no results: query_len=%d", len(query))
         return results
--- a/src/soul_protocol/runtime/types.py
+++ b/src/soul_protocol/runtime/types.py
@@ -276,6 +276,15 @@ class MemoryEntry(BaseModel):
     # v0.4.0 — Contradiction detection
     superseded: bool = False  # True when a newer memory contradicts this one
     visibility: MemoryVisibility = MemoryVisibility.BONDED
+    # F2 archival memory — marks episodic memories that have been archived
+    archived: bool = False  # True when memory has been compressed into a ConversationArchive
+    # F1 progressive disclosure — runtime-only marker, never persisted
+    is_summarized: bool = False  # Runtime marker: True when content replaced with abstract
+    # Move 5 PR-A — RBAC/ABAC scope tags. Empty list = no scope assigned
+    # (visible to any caller). Hierarchical glob: "org:sales:*" matches
+    # "org:sales:leads". Filtered at retrieval time before results reach
+    # the LLM.
+    scope: list[str] = Field(default_factory=list)
 
 
 class CoreMemory(BaseModel):
--- a/src/soul_protocol/spec/__init__.py
+++ b/src/soul_protocol/spec/__init__.py
@@ -49,6 +49,7 @@ from .decisions import (
 from .journal import ACTION_NAMESPACES, Actor, DataRef, EventEntry
 from .manifest import Manifest
 from .memory import DictMemoryStore, Interaction, MemoryEntry, MemoryStore, MemoryVisibility, Participant
+from .scope import match_scope, normalise_scopes
 from .template import SoulTemplate
 from .learning import LearningEvent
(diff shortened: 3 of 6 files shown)


prakashUXtech (commented):
## Review — feat(spec): scope tags on MemoryEntry + match_scope helper

**Verdict: Blockers to resolve before merge.** The core design is sound — fail-open semantics for back-compat are the right default for a field being rolled out incrementally. But two issues need to settle before this goes into the wild in `.soul` files: an ambiguous grammar edge case in `match_scope` and a missing normalization call on the `scopes=` kwarg at the recall layer.

---

### What's good

The fail-open / fail-closed contract is well-specified and tested:
- Memory without scope: always visible (back-compat for all existing souls).
- Caller without scope filter: sees everything (unchanged retrieval for code that doesn't opt in).
- Memory + caller both scoped: intersection decides.

This is the only design that lets you roll out scope tagging incrementally without a flag day. The decision is correct and the test suite proves each branch.

`normalise_scopes` is the right utility to ship alongside the grammar. Centralized lowercasing and deduplication prevents the most common tag divergence (casing mismatches between producer and consumer).

---

### Findings

**1 — Blocker: `match_scope` in `soul.recall` applies to the raw caller-supplied list, not the normalized form.**

`runtime/soul.py`, the scope filter added by this PR:

```python
if scopes:
    from soul_protocol.spec.scope import match_scope
    results = [
        entry for entry in results
        if match_scope(getattr(entry, "scope", None), scopes)
    ]
```

`normalise_scopes` is exported but never called here. A caller passing `["Org:Sales:*"]` will fail to match memories tagged `["org:sales:leads"]` because `match_scope` is case-sensitive by design (the docstring says so), but `normalise_scopes` lowercases inputs. The two utilities exist in the same module and are meant to work together — if the recall path does not normalize the caller's scopes before matching, the guarantee in the PR description ("Comparison is case-sensitive — scope tags are namespaced ASCII identifiers") is correct for the helper in isolation but misleading for callers who read the PR and expect case-insensitive matching at the recall layer.

Fix:

```python
if scopes:
    from soul_protocol.spec.scope import match_scope, normalise_scopes
    normalised = normalise_scopes(scopes)
    results = [
        entry for entry in results
        if match_scope(getattr(entry, "scope", None), normalised)
    ]
```

Alternatively, document clearly in `recall`'s docstring that scope tags are matched case-sensitively and callers must normalize before passing. Given that the PR adds `normalise_scopes` to the public spec API, the first option is cleaner.

**2 — Protocol-stability concern: `match_scope("org:sales", ["org:sales:*"])` returns True — but is that intended?**

`spec/scope.py`, `_granted`:

```python
if allowed.endswith(":*"):
    prefix = allowed[:-2]
    return entity == prefix or entity.startswith(prefix + ":")
```

The `entity == prefix` branch means `org:sales` is granted by `org:sales:*`. The test suite explicitly validates this (`test_glob_matches_exact_prefix`: "The bare prefix itself is included"). This is a reasonable choice but it needs to be locked in the spec docstring because it has non-obvious access control consequences: a memory tagged `org:sales` (the parent scope) is visible to anyone holding `org:sales:*` (a child wildcard). In most RBAC systems the inverse is true — holding a child scope doesn't grant the parent.

This is not a bug given the current tests, but before souls with these tags are in the wild, confirm intentionality and add a sentence to the `match_scope` docstring: "A wildcard `org:sales:*` also grants access to the bare prefix `org:sales`." That way future maintainers don't silently revert it assuming it's a bug.

---

### Test gaps

The recall filtering tests manipulate `soul._memory._semantic._facts` directly to inject scoped entries. That's fine for now, but there's no test for the normalization gap (point 1 above) — adding a case where the caller passes mixed-case scopes and checking that recall returns the right entries would catch any future regression when the fix lands.

Also missing: what happens when a memory's `scope` list contains an entry that was stored before `normalise_scopes` existed — i.e., stored as `"Org:Sales:*"` (capitalized). The `getattr(entry, "scope", None)` path in recall passes the raw stored value to `match_scope` without normalization. If memories can be written with un-normalized scopes (nothing prevents it at store time today), recall can silently drop results that should match. Worth a test and a call to normalize both sides.

---

### Nit

`spec/scope.py` documents that comparison is case-sensitive, but `normalise_scopes` lowercases. The two are inconsistent if only one side is normalized. The docstring on `match_scope` should say "both sides should be normalized via `normalise_scopes` before calling" rather than leaving this as implicit convention.

---

**Targets `main` — should be against `dev`. Given the RBAC implications of this primitive, confirm target branch before merge.**

PR #163, opened 2026-04-13
feat(templates): bundled role archetypes + load_template + soul template CLI

prakashUXtech:
Adds the role-template library that lets a non-technical operator create a domain-specific soul without authoring OCEAN traits or core memories from scratch. Four bundled archetypes ship with the package; \`SoulFactory.load_template()\` reads any YAML/JSON file so custom templates live anywhere.

Move 6 PR-A. PR-B adds the picker UI in paw-enterprise.

### \`src/soul_protocol/templates/\` (new)
- **arrow.yaml** — sales companion. High conscientiousness, moderate extraversion, low neuroticism. Default scope \`org:sales:*\`. Recommended tools: \`instinct_propose\`, \`instinct_corrections\`, \`fabric_query\`, \`kb_search\`.
- **flash.yaml** — content companion. High openness + agreeableness for tone matching. Scope \`org:marketing:*\`.
- **cyborg.yaml** — recruiting companion. Very low neuroticism for interview cadence reliability. Scope \`org:hr:recruiting\`.
- **analyst.yaml** — research companion. Explicit citation discipline. Scope \`org:research:*\`.
- \`__init__.py\` exports \`BUNDLED_TEMPLATES\`, \`template_path(name)\`, \`list_bundled()\` so paw-runtime + the CLI agree on naming.

### \`src/soul_protocol/runtime/templates.py\`
- \`SoulFactory.load_template(path)\` — YAML + JSON loader, extension-driven.
- \`SoulFactory.load_bundled(name)\` — shorthand for the bundled set.
- \`SoulFactory.get(name)\` — accessor for registered templates.

### CLI
- \`soul template list\` — every bundled template with archetype + first three skills.
- \`soul template show <name>\` — raw YAML for one.
- \`soul create --template <name> [--name X] [--dir .soul] [--format dir|zip]\` — instantiates a soul from a bundled template.

### \`pyproject.toml\`
- Force-include the templates directory so YAML files ship in the wheel.

## Test plan

- [x] 16 new tests in \`tests/test_bundled_templates.py\` covering bundled inventory contracts (every named template loads + validates, OCEAN values 0-1, default scope on Arrow, recommended tools on Cyborg, every template ships with at least one skill), YAML + JSON loaders, missing-file behaviour, end-to-end Soul instantiation, helpers.
- [x] Existing \`test_soul_templates.py\` (33 tests covering \`from_template\` + \`batch_spawn\` + registry) untouched
- [x] \`pytest tests/\` — 2121 passed
- [x] \`ruff check\` — clean

## Pairs with

- soul-protocol [#162](https://github.com/qbtrix/soul-protocol/pull/162) — scope tags (templates carry a recommended \`default_scope\` in metadata)
- PR-B (paw-enterprise) — picker UI for the agent + pocket creation flows

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,6 +84,9 @@ Issues = "https://github.com/qbtrix/soul-protocol/issues"
 [tool.hatch.build.targets.wheel]
 packages = ["src/soul_protocol"]
 
+[tool.hatch.build.targets.wheel.force-include]
+"src/soul_protocol/templates" = "soul_protocol/templates"
+
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
 testpaths = ["tests"]
--- a/src/soul_protocol/cli/main.py
+++ b/src/soul_protocol/cli/main.py
@@ -674,6 +674,104 @@ def list():
     asyncio.run(_list())
 
 
+@cli.group()
+def template():
+    """Browse and instantiate bundled role templates (Move 6)."""
+
+
+@template.command("list")
+def template_list():
+    """Show every bundled template (Arrow, Flash, Cyborg, Analyst, ...)."""
+    from soul_protocol.templates import list_bundled
+
+    names = list_bundled()
+    if not names:
+        console.print("[dim]No bundled templates installed[/dim]")
+        return
+
+    table = Table(title="Bundled Soul Templates", border_style="blue")
+    table.add_column("Name", style="cyan")
+    table.add_column("Archetype")
+    table.add_column("Skills")
+
+    from soul_protocol.runtime.templates import SoulFactory
+
+    for name in sorted(names):
+        try:
+            tmpl = SoulFactory.load_bundled(name)
+            table.add_row(
+                tmpl.name,
+                tmpl.archetype,
+                ", ".join(tmpl.skills[:3]) + ("..." if len(tmpl.skills) > 3 else ""),
+            )
+        except Exception as exc:
+            table.add_row(name, "[red]load error[/red]", str(exc))
+
+    console.print(table)
+
+
+@template.command("show")
+@click.argument("name")
+def template_show(name):
+    """Print the YAML for a single bundled template."""
+    from soul_protocol.templates import template_path
+
+    p = template_path(name)
+    if not p.exists():
+        console.print(f"[red]No bundled template named '{name}'[/red]")
+        raise SystemExit(1)
+    console.print(p.read_text(encoding="utf-8"))
+
+
+@cli.command("create")
+@click.option("--template", "template_name", required=True, help="Bundled template name")
+@click.option("--name", help="Override the soul name")
+@click.option(
+    "--dir",
+    "soul_dir",
+    default=".soul",
+    help="Directory to write the new soul to (default: .soul)",
+)
+@click.option(
+    "--format",
+    "soul_format",
+    type=click.Choice(["dir", "zip"], case_sensitive=False),
+    default="dir",
+    help="Storage format",
+)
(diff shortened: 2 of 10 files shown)


prakashUXtech (commented):
## Review — feat(templates): bundled role archetypes + load_template + soul template CLI

**Verdict: LGTM with one minor issue and a nit on error handling.** The template library is well-constructed. YAML files ship with the wheel via the `force-include` fix, the loader is extension-driven with a clear error when PyYAML is absent, the CLI commands are discoverable and idempotent, and the 16 new tests cover the contracts that matter most (every bundled template loads and validates, OCEAN values in range, missing file raises).

---

### What's good

The `force-include` entry in `pyproject.toml` is the right fix for shipping YAML data files with a wheel — this is the exact problem that silently breaks template loading in installed packages.

`load_bundled` using `template_path(name)` from the `templates` package keeps the naming contract in one place. `list_bundled()` reading from disk rather than `BUNDLED_TEMPLATES` is the right choice — it means a user who drops a custom YAML into the package directory automatically shows up in `soul template list`.

The `soul create` command is clean: no collision with the existing `birth` or `init` commands. The path handling (writing to `--dir` as a zip vs. directory depending on `--format`) is consistent with the existing export behavior.

---

### Findings

**1 — Minor: `create_from_template` returns 1 on unknown template but the exit path may swallow it.**

`cli/main.py`, the `create_from_template` function:

```python
async def _go():
    try:
        tmpl = SoulFactory.load_bundled(template_name)
    except FileNotFoundError:
        console.print(f"[red]Unknown template: {template_name}[/red]")
        ...
        return 1
    ...
    return 0

raise SystemExit(asyncio.run(_go()) or 0)
```

`asyncio.run(_go()) or 0` — if `_go()` returns `1`, `1 or 0` evaluates to `1`, so `SystemExit(1)` is raised correctly. But if `_go()` returns `None` (which it would if an unexpected exception propagates past the `try/except`), `None or 0` evaluates to `0` and the command exits clean on failure. The existing pattern in this file uses bare `asyncio.run()` without the `or 0` guard. Wrapping in an explicit `if` makes the contract clearer and doesn't change behavior for the happy path:

```python
raise SystemExit(asyncio.run(_go()))
```

**2 — Nit: `default_scope` lives in `metadata` (untyped dict) rather than a typed field.**

The template YAML files carry `metadata.default_scope` and `metadata.recommended_tools`. These are consumed by the CLI and tests via `tmpl.metadata.get("default_scope", [])`. Because `SoulTemplate.metadata` is `dict[str, Any]`, there's no validation that `default_scope` is a list of strings or that `recommended_tools` is a list of tool names. A misspelled key in a custom template fails silently at load time and only surfaces when paw-runtime tries to apply the scope.

This is a v1 tradeoff that's acceptable today, but it should be tracked. As the `default_scope` field is directly connected to the scope grammar from PR #162, consider promoting it to a typed field on `SoulTemplate` in a follow-up:

```python
default_scope: list[str] = Field(default_factory=list)
recommended_tools: list[str] = Field(default_factory=list)
```

That would let paw-runtime read `tmpl.default_scope` instead of `tmpl.metadata.get("default_scope", [])` and get Pydantic validation for free.

---

### Test gaps

No test for `soul template show` with an unknown name. The command does a `p.exists()` check and calls `raise SystemExit(1)` — a quick CLI runner test would lock the exit code contract. Also no test for the `--format zip` branch of `soul create` (the directory branch is covered implicitly by `test_from_template_creates_soul_with_archetype`). Both are minor and fine to add in PR-B.

---

**Targets `main` — confirm whether `dev` is the intended merge target before landing.**

PR #164, opened 2026-04-13
docs: org journal spec + architecture trim

prakashUXtech:
## Summary

Ships the framework-agnostic org-layer protocol spec for Soul Protocol, and trims the existing architecture doc of redundant Paw-specific content.

## What's in the PR

**New doc:** `docs/org-journal-spec.md` — protocol-pure RFC covering:
- Org Journal (`EventEntry`, `Actor`, `DataRef` + invariants + action namespace catalog)
- Root Agent concept (governance identity, three-layer undeletability, no conversational surface)
- Soul hierarchy (root / team / user / agent / pocket-as-scope)
- RetrievalRouter + CredentialBroker contracts
- DataRef Zero-Copy payload variant + tradeoff table + non-hermetic replay semantics
- Decision Traces (`agent.proposed` + `human.corrected` + `decision.graduated`)
- Storage tiers, security model, open questions, conformance notes for implementers

**Updated doc:** `docs/architecture.md` — stripped redundant sections that were moved to the spec. Section 7 now a neutral implementation summary for *this codebase*, flagging four known cosmetic leaks the cleanup PR will address.

**New doc:** `docs/manual-testing.md` (earlier in this stack) — banner added noting the CLI rename pending.

## Why the scope changed

Original RFC (`docs/org-architecture.md`) mixed framework-agnostic protocol semantics with Paw OS-specific migration planning and implementation details. Review flagged this as a positioning leak — anyone evaluating soul-protocol for non-Paw use would see Paw branding throughout. Split into:

- **This PR:** protocol spec, framework-agnostic
- **Paw-specific migration plan:** moved to `paw-workspace/MIGRATION.md` (workspace-level, not in this repo)

## Known cosmetic leaks (follow-up cleanup PR)

The spec is clean but the codebase still has four leaks that need fixing before v0.3 merges to dev:

1. CLI `soul paw os *` → should be `soul org *`
2. Default data dir `~/.pocketpaw/` → should be `~/.soul/` with `SOUL_DATA_DIR` env override
3. Action namespace `paw.os.destroyed` → should be `org.destroyed`
4. Docstrings + help text reference "Paw OS" → should reference "org"

Separate PR will handle these against the v0.3 stack branches.

## Test plan

- [ ] Captain reads protocol spec end-to-end
- [ ] architect agent review (internal consistency, framework-agnostic verification)
- [ ] Reconcile open questions before the cleanup PR ships

PR #165, opened 2026-04-13
feat(spec): journal EventEntry + Actor + DataRef

prakashUXtech:
First slice of the Org Architecture RFC (#164). Ships the spec-only Pydantic models for the org journal — the append-only, UTC-stamped, scope-tagged event log the RFC proposes as Paw OS's source of truth.

## What's in

- `Actor` — `kind` (agent/user/system/root), `id`, `scope_context`. No anonymous writes.
- `DataRef` — `source`, `query`, `point_in_time` (tz-aware enforced), `cache_policy`, `cache_ttl_s`. The reference type for zero-copy retrieval against live systems.
- `EventEntry` — `id`, `ts` (tz-aware enforced), `actor`, `action`, `scope` (required + non-empty), `causation_id`, `correlation_id`, `payload: DataRef | dict`, optional `prev_hash` and `sig` as bytes that round-trip through JSON as base64.
- `ACTION_NAMESPACES` — initial catalog from the RFC appendix, kept as a tuple rather than an enum so verticals can ship new actions without a library bump.

Semantics locked in docstrings: append-only, scope required, UTC-only timestamps, system:* actors reserved for subsystems.

## What's not

No engine code, no CLI, no migration scripts. The SQLite WAL engine and `paw os init` come in follow-up PRs. This PR is deliberately small so the shape of the models can get reviewed before anything depends on them.

## Tests

`tests/test_spec/test_journal.py` — 21 tests covering:
- JSON round-trips for each model
- `ts` and `point_in_time` reject naive datetimes (the naive-datetime bugs from today's review won't recur here)
- `scope` is required and every entry must be non-empty
- `payload` union round-trips for both `dict` and `DataRef` variants
- `Actor.kind` rejects unknown values; `id` must be non-empty
- `causation_id` is `None` for genesis events and round-trips when set
- `ACTION_NAMESPACES` covers the catalog, and `action` itself stays free-form

All 62 tests in `tests/test_spec/` still pass.

## Design source

Matches the shape laid out in #164 § "The Org Journal" and the Appendix event catalog. Same spec-only PR shape as #161 (RetrievalTrace) and #162 (scope tags).

### DIFF
--- a/src/soul_protocol/__init__.py
+++ b/src/soul_protocol/__init__.py
@@ -16,6 +16,8 @@
 #   v0.2.1 — Added CognitiveEngine, HeuristicEngine, ReflectionResult exports.
 #   v0.2.0 — Added psychology types (SomaticMarker, SignificanceScore,
 #   GeneralEvent, SelfImage) to public exports.
+# Updated: feat/journal-spec — Added org journal primitives (Actor, DataRef,
+#   EventEntry, ACTION_NAMESPACES) from spec.journal. See RFC PR #164.
 
 from __future__ import annotations
 
@@ -68,6 +70,7 @@ from .runtime.eternal import ArchiveResult, EternalStorageManager, EternalStorag
 # Core primitives from spec/ (always available — only requires pydantic)
 from .spec.identity import BondTarget as CoreBondTarget
 from .spec.identity import Identity as CoreIdentity
+from .spec.journal import ACTION_NAMESPACES, Actor, DataRef, EventEntry
 from .spec.manifest import Manifest as CoreManifest
 from .spec.memory import DictMemoryStore, MemoryStore
 from .spec.memory import Interaction as CoreInteraction
@@ -138,6 +141,11 @@ __all__ = [
     "CoreParticipant",
     "DictMemoryStore",
     "MemoryStore",
+    # Org Journal (feat/journal-spec)
+    "ACTION_NAMESPACES",
+    "Actor",
+    "DataRef",
+    "EventEntry",
 ]
 
 __version__ = "0.2.5"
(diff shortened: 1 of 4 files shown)


prakashUXtech (commented):
**Verdict:** LGTM with one spec concern and one nit

**Top findings**

1. **Spec gap: DataRef allows non-UTC timezone-aware datetimes.** The validator on `point_in_time` accepts any tz-aware datetime, and `test_dataref_accepts_non_utc_tz_but_preserves_offset` explicitly documents this as intentional. The spec in #164 says "UTC everywhere" as an invariant. If a non-UTC-offset `point_in_time` ends up in the journal, replays and external consumers have to normalize on read rather than on write. Recommend either enforcing UTC normalization in the validator (`.astimezone(UTC)` and return) or adding a `utcoffset() == timedelta(0)` check that raises, matching how `ts` is handled on `EventEntry`. The current behavior is a quiet inconsistency between the two datetime fields in the same module.

2. **Deserialization ambiguity on `payload: DataRef | dict`.** Pydantic v2 resolves union arms left-to-right in "smart union" mode, but when deserializing a JSON object that happens to have keys matching DataRef fields (`source`, `query`, `point_in_time`), Pydantic tries DataRef first and will succeed if the fields validate. The round-trip test catches the clean case; an adversarial dict payload with a `"source"` key could silently deserialize as a DataRef. This matters once the journal engine serializes and replays events from storage. Consider adding a tagged discriminator field (e.g. `type: Literal["dataref"]` on DataRef) or reversing the union to `dict | DataRef` and adding a model validator that upgrades to DataRef when `"type": "dataref"` is present. Not a blocker for the spec slice, but worth resolving before the engine layer lands and payloads are persisted.

**Nits**

- `test_dataref_accepts_non_utc_tz_but_preserves_offset` includes a comment "callers are expected to normalize to UTC" but the validator does not enforce or perform that normalization. The comment sets an expectation the code does not meet. Update the comment to match reality or implement normalization.
- `ACTION_NAMESPACES` omits `fabric.object.created`, `fabric.object.updated`, and `fabric.object.archived` that appear in the spec appendix in #164. Either add them here or add a note explaining the omission is deliberate. The tuple is a discoverability aid; partial coverage creates drift.

**Positive notes**

The `JournalBytes` type alias is clean — round-tripping `bytes` through JSON as base64 via Pydantic's `PlainSerializer`/`BeforeValidator` composition is the right approach and avoids ad-hoc encoding scattered across the codebase. The test suite is thorough on the happy path and the most important invariant (naive datetime rejection). 21 tests for a 216-line module is a good ratio.

PR #166, opened 2026-04-13
feat(engine): journal SQLite WAL backend

prakashUXtech:
Stacks on #165 — the spec models (`EventEntry`, `Actor`, `DataRef`, `ACTION_NAMESPACES`) need to land first. Implements slice 2 of Workstream A from the Org Architecture RFC (#164).

## What this adds

`src/soul_protocol/engine/journal/` — the append-only journal engine:

- `Journal` — the high-level class. Owns the invariants: timezone-aware UTC on every `ts`, non-empty scope, attributed actor, monotonic `ts` across the journal, auto-assigned `seq`. Any caller `prev_hash` is preserved; otherwise the engine computes a sha256 link over the previous event opportunistically.
- `JournalBackend` — the Protocol. One method does seq-assignment and insertion atomically so concurrent writers can't race the sequence number.
- `SQLiteJournalBackend` — the v1 implementation. `sqlite3` stdlib, WAL mode on open, `BEGIN IMMEDIATE` transactions, a threading lock so a single backend can be shared across threads.
- `schema.py` — SQL + migration helper. First write brings an empty DB to v1; the meta table tracks the version for future migrations.
- `exceptions.py` — `JournalError`, `SchemaError`, `IntegrityError`, `NotFoundError`.
- `open_journal(path)` — convenience factory.

## Scope filter

Prefix-plus-wildcard matching with per-segment arity rules:

- `org:sales` matches exactly `org:sales`.
- `org:sales:*` matches `org:sales:leads` but not bare `org:sales`.
- `org:*:leads` matches `org:sales:leads` and `org:hr:leads`.
- `org:*` matches `org:sales` but not bare `org`.

This mirrors the arity semantics called out in the RFC. The richer grammar from #162 (`spec.scope`) will replace this once that module lands — the internal `_scope_matches` helper is the seam.

## Deferred to later slices

- CLI (`paw os init`, `paw journal append/query`), FastAPI surface, onboarding wizard.
- Required hash-chain + event signing — `prev_hash` is nullable today, computed best-effort.
- Admin-only action enforcement.
- Async wrappers — the journal is sync by design; consumers can wrap.

## Tests

20 tests in `tests/test_engine/test_journal.py`. Round-trip, every filter dimension, combined filters, scope wildcard matcher unit + integration, tz rejection on append and query, monotonic ts, seq gap-free ordering, replay cursor, `DataRef` payload survives through SQLite, two threaded writers against a shared WAL DB without data loss, zero-to-v1 migration, backend protocol conformance.

Engine + tests total 873 LOC, a touch over the 800 target. The concurrency test accounts for roughly 60 of those on its own.

## Verification

- `uv run pytest tests/test_engine/test_journal.py -v` — 20 passed
- `uv run pytest tests/` — 1952 passed

## References

- RFC: #164
- Slice 1 (spec models): #165

### DIFF
--- a/src/soul_protocol/engine/__init__.py
+++ b/src/soul_protocol/engine/__init__.py
@@ -1,0 +1,4 @@
+# __init__.py — soul_protocol.engine package.
+# Created: feat/journal-engine — Workstream A slice 2 of Org Architecture RFC (#164).
+# Engine-layer modules implement the runtime primitives (journal, fabric, …)
+# behind the spec models in `soul_protocol.spec`.
--- a/src/soul_protocol/engine/journal/__init__.py
+++ b/src/soul_protocol/engine/journal/__init__.py
@@ -1,0 +1,21 @@
+# __init__.py — Public API for the journal engine.
+# Created: feat/journal-engine — Workstream A slice 2 of Org Architecture RFC (#164).
+# Re-exports Journal, JournalBackend, open_journal, and the exception hierarchy.
+
+from __future__ import annotations
+
+from .backend import JournalBackend
+from .exceptions import IntegrityError, JournalError, NotFoundError, SchemaError
+from .journal import Journal, open_journal
+from .sqlite import SQLiteJournalBackend
+
+__all__ = [
+    "IntegrityError",
+    "Journal",
+    "JournalBackend",
+    "JournalError",
+    "NotFoundError",
+    "SQLiteJournalBackend",
+    "SchemaError",
+    "open_journal",
+]
--- a/src/soul_protocol/engine/journal/backend.py
+++ b/src/soul_protocol/engine/journal/backend.py
@@ -1,0 +1,45 @@
+# backend.py — Journal backend Protocol.
+# Created: feat/journal-engine — Workstream A slice 2 of Org Architecture RFC (#164).
+# The Journal class delegates persistence to a JournalBackend. v1 ships a
+# single SQLite implementation; future backends (LMDB, Postgres, pluggable
+# remote) implement the same Protocol.
+
+from __future__ import annotations
+
+from collections.abc import Iterator
+from datetime import datetime
+from typing import Protocol, runtime_checkable
+from uuid import UUID
+
+from soul_protocol.spec.journal import Actor, EventEntry
+
+
+@runtime_checkable
+class JournalBackend(Protocol):
+    """Storage contract for journal events. All methods are synchronous."""
+
+    def append(self, entry: EventEntry) -> int:
+        """Persist `entry`, assigning and returning its monotonic seq atomically."""
+
+    def query(
+        self,
+        *,
+        action: str | None = None,
+        actor: Actor | None = None,
+        scope: list[str] | None = None,
+        correlation_id: UUID | None = None,
+        since: datetime | None = None,
+        until: datetime | None = None,
+        limit: int = 100,
+        offset: int = 0,
+    ) -> list[EventEntry]:
+        """Return events matching the conjunction of filters."""
+
+    def replay_from(self, seq: int = 0) -> Iterator[EventEntry]:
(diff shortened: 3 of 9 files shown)


prakashUXtech (commented):
**Verdict:** Approve with one blocking fix and two important follow-ups before merging to dev.

---

**Top findings**

**1. [Blocking] ts monotonicity check is not thread-safe across two Journal instances on the same DB**

`Journal.append()` calls `self._backend.last_entry()` outside the `BEGIN IMMEDIATE` transaction. When two writers share the same SQLite file (different `Journal` instances, as the WAL test exercises), the sequence is:

1. Writer A reads `last_entry()` — gets ts T.
2. Writer B reads `last_entry()` — also gets ts T (B hasn't committed yet).
3. Writer A writes event at ts T+1, commits.
4. Writer B writes event at ts T-1 (its own clock), passes the pre-transaction check, hits `BEGIN IMMEDIATE`, and the write lands.

The per-instance threading lock (`self._lock`) protects one journal object but not cross-instance access. The fix is to either move `last_entry()` inside the `BEGIN IMMEDIATE` block in `SQLiteJournalBackend.append()`, pushing the monotonicity check into the backend where it can be atomic, or to add a DB-level `CHECK` constraint on `ts`. The concurrent test passes today because the shared `tick()` fixture guarantees global monotonicity — it doesn't catch the gap.

**2. [Important] Silent hash chain corruption on mid-chain write failure**

In `Journal.append()`, when `_hash_link` succeeds but the subsequent `_backend.append()` raises, the entry object was already mutated with the new `prev_hash` via `model_copy`. The next call will see the empty journal (the failed write was rolled back), fetch `None` from `last_entry()`, skip hash linkage, and record an event with no `prev_hash`. The chain is silently broken with no signal to consumers. Consider at minimum logging the exception from the bare `except Exception: pass` block on the hash step, and separately on the backend write failure, so operators have some visibility.

**3. [Important] Scope filter applied in Python after fetching all rows**

In `SQLiteJournalBackend.query()`, `action`, `actor`, `correlation_id`, and time window all push predicates into SQL, but `scope` is applied as a Python post-filter after `fetchall()`. On a large journal, a query like `journal.query(scope=["org:sales:*"])` with no other filters loads every row into memory. The `scope` column is already stored as JSON and indexed by `seq`; adding a `scope` index and pushing at least exact-match scope filtering into SQL would prevent this. Wildcard patterns would still need Python filtering, but exact matches could short-circuit earlier.

---

**Test gaps**

- The concurrent test's shared `tick()` clock masks the cross-instance monotonicity race. Worth adding a test where two journals write with independent `datetime.now(UTC)` calls and asserting the DB still has no out-of-order ts.
- No test for `close()` idempotency — the implementation handles it (`ProgrammingError` catch), but there is no test confirming a second `close()` does not raise.
- No test for what happens when `begin_immediate` itself fails (e.g., lock timeout). The `ROLLBACK` in the `except` block runs on the same connection that timed out; that behavior deserves at least a comment explaining why it is safe.

---

**Nits**

- `open_journal` docstring says "and migrate on first write" but migration runs at `__init__` time, not on first write. Minor wording issue.
- The `migrate()` function commits inside itself but `SQLiteJournalBackend.__init__` is using `isolation_level=None` (autocommit). The `conn.commit()` at the end of `migrate()` is a no-op in that mode. Not harmful but confusing.
- `replay_from` on a closed backend will hit `ProgrammingError` without a clear error message — same idempotency gap as `close()`.

---

**Positive notes**

The Protocol shape for `JournalBackend` is clean — one method per responsibility, explicit return types, no ABC inheritance tax. The `BEGIN IMMEDIATE` choice is the right call for SQLite WAL: it blocks other writers at transaction start rather than at first write, which is what you want for seq assignment. The test coverage for filter dimensions is thorough and readable. The `DataRef` round-trip test and the zero-to-v1 migration test are exactly the kind of regression anchors this layer needs.

PR #167, opened 2026-04-13
feat(cli): soul org init — bootstrap org with root agent + genesis event

prakashUXtech:
## Summary

Ships `soul paw os init`, the onboarding command that turns an empty directory into a working Paw OS org. Part of Workstream A slice 3 of the org architecture RFC (#164). Stacks on top of #166 (journal engine).

Running it once with `--org-name "Acme Ventures" --purpose "A software company"` produces:

```
~/.pocketpaw/org/
├── root.soul              # Governance soul, zip-packed
├── journal.db             # SQLite WAL journal, 2 genesis events
└── keys/
    ├── root.ed25519       # 0600 private key
    ├── root.ed25519.pub   # raw public key
    └── root.did           # the root DID string
```

The journal starts with an `org.created` event followed by a `scope.created` event for `org:*`, and the second event names the first as its `causation_id`. No signing yet — that lands in a later slice.

## Design calls

- The governance persona is hardcoded for now. The RFC plans to load it via `load_template("governance")` from slice #163, but that loader isn't in the installed package yet, so there is a TODO at the fallback pointing to where it swaps in.
- Ed25519 key generation uses `cryptography`, which is already pulled in through the `engine` extra. No new dependency.
- Command is nested as `soul paw os init` rather than a top-level `paw` entrypoint — keeps the existing `soul` console script as the single install surface while still making room for `soul paw os status` and `soul paw os destroy` later.
- Running twice without `--force` exits non-zero with a clear message. `--force` wipes the directory in-place so the user can iterate without chasing removal commands.

## Scope

- New CLI module: `src/soul_protocol/cli/paw_os.py`
- Register the subgroup in `src/soul_protocol/cli/main.py`
- 8 tests in `tests/test_cli/test_paw_os_init.py` covering the happy path, idempotency, `--force`, `--non-interactive` guard, journal queryability, 0600 permissions on the key file, `Soul.awaken()` round-trip, and a sanity check that the private key never leaks into a journal payload
- Docs: new `docs/paw-os.md` and a pointer from `docs/getting-started.md`

## Test plan

- [x] `uv run pytest tests/test_cli/test_paw_os_init.py -v` — 8 pass
- [x] `uv run pytest tests/` — full suite 1979 pass
- [x] Manual: `soul paw os init --org-name "Smoke" --data-dir /tmp/org --non-interactive` and inspect the journal


prakashUXtech (commented):
**Verdict:** Blockers — docs need a full rewrite before merge

**Summary**

The CLI and runtime code in this PR are clean. The rename from \`soul paw os\` to \`soul org\` landed correctly in \`cli/org.py\`, \`cli/main.py\`, and the test file. The problem is squarely in the docs, where the rename did not happen at all.

---

**Paw-leak audit: 9 leaks across 2 files**

\`docs/paw-os.md\` (all 9 leaks are user-facing and public):

- Line 1 (HTML comment): \`Paw OS onboarding — \`soul paw os init\`\`
- Line 4: heading \`# Paw OS\`
- Line 6: body copy references "Paw OS org" twice and "Paw OS instance"
- Line 8: subsection heading \`## \`soul paw os init\`\`
- Line 13: code block shows \`soul paw os init\` — wrong command
- Line 20: default path shown as \`~/.pocketpaw/org/\` — should be \`~/.soul/\`
- Lines 26, 29: filesystem tree diagram shows \`~/.pocketpaw/org/\`
- Line 47: "soul paw os fleet install" — wrong command

\`docs/getting-started.md\` line 279: link text says "Paw OS" and the inline command shows \`soul paw os init\`.

The doc page is essentially a pre-rename copy-paste. Any user who reads it will run the wrong commands against a directory that no longer exists.

---

**Changes requested**

1. Rename \`docs/paw-os.md\` to \`docs/org.md\` (or keep the filename but clear the content). Update every instance of "Paw OS" to "org", every \`soul paw os\` invocation to \`soul org\`, and both path references from \`~/.pocketpaw/org/\` to \`~/.soul/\`.

2. Fix \`docs/getting-started.md\` line 279 — link text and inline command both need updating to match.

---

**What's good**

- Ed25519 key write using \`os.open(..., 0o600)\` before \`os.write()\` is the right approach — no window where group/other bits are set.
- \`try/finally: journal.close()\` around all appends is solid — journal handle always released even if an append raises.
- The private-key-not-in-journal-payloads test is a nice explicit safety net.
- 8 tests with good coverage of the non-interactive path, idempotency, \`--force\`, and permission check.

---

**Nits (non-blocking)**

- If \`_build_governance_soul\` raises between key write and journal init, the org dir is left with a root soul and keys but no journal — rerunning without \`--force\` then fails with "already exists". A note in the help text about \`--force\` as the recovery path would help, or a brief mention in docs.
- \`_remove_tree\` swallows all errors silently via the caller. Worth at least a \`console.print(warning)\` if a child cannot be removed.

prakashUXtech (commented):
test ping

prakashUXtech (commented):
**Verdict:** Blockers — docs need a full rewrite before merge

**Summary**

The CLI and runtime code in this PR are clean. The rename from \`soul paw os\` to \`soul org\` landed correctly in \`cli/org.py\`, \`cli/main.py\`, and the test file. The problem is squarely in the docs, where the rename did not happen at all.

---

**Paw-leak audit: 9 leaks across 2 files**

\`docs/paw-os.md\` — all user-facing:

- Line 1 (HTML comment): references \`soul paw os init\`
- Line 4: heading is \`# Paw OS\` — should be \`# Org\` (or \`# Soul Org Init\`)
- Line 6: body copy says "Paw OS org" and "Paw OS instance" throughout
- Line 8: subsection heading shows \`soul paw os init\`
- Line 13: code block runs \`soul paw os init\` — will fail for any user who copies it
- Line 20: default path shown as \`~/.pocketpaw/org/\` — actual default is \`~/.soul/\`
- Lines 26–29: filesystem tree diagram shows \`~/.pocketpaw/org/\`
- Line 47: next-step command is \`soul paw os fleet install\` — wrong group

\`docs/getting-started.md\` line 279: link text says "Paw OS" and inline command shows \`soul paw os init\`.

The doc page is a pre-rename snapshot. Any reader who follows it will run the wrong command against a path that no longer exists.

---

**Changes requested**

1. Rename or rewrite \`docs/paw-os.md\`: replace every "Paw OS" with "Org", every \`soul paw os\` invocation with \`soul org\`, and both path references from \`~/.pocketpaw/org/\` to \`~/.soul/\`.

2. Fix \`docs/getting-started.md\` line 279 — link text and inline command both need updating.

---

**What's good**

- Ed25519 key write using \`os.open(..., 0o600)\` before \`os.write()\` is the correct approach — no window where group/other bits are set.
- \`try/finally: journal.close()\` around all appends is solid.
- The private-key-not-in-journal-payloads test is a nice explicit safety net.
- 8 tests covering the non-interactive path, idempotency, \`--force\`, and permission check — good baseline.

---

**Nits (non-blocking)**

- If anything between key write and \`journal.append\` raises, the org dir is left with a soul and keys but no journal. Rerunning without \`--force\` then fails with "already exists". A one-liner in the CLI output pointing to \`--force\` as the recovery path would help.
- \`_remove_tree\` silently swallows errors on individual children — worth at least a console warning if a child cannot be removed.

PR #168, opened 2026-04-13
feat(spec): decision traces — agent.proposed + human.corrected payload types

prakashUXtech:
## What this is

The spec side of Workstream D from RFC #164 — the decision-trace event chain.

Every agent proposal a reviewer edits, accepts, rejects, or defers now has a structured payload type that rides inside a journal `EventEntry`. Over time, recurring corrections can be clustered, and the pattern promoted from raw events into standing guidance the agent loads as memory. That's the loop the gaps doc calls out as the moat: Decisions, as a first-class data type, that neither SAP nor Salesforce are built to carry.

This PR ships only the primitives. The pocketpaw-side emit points — tool-call preview, draft-approval UI hooks, the edit-captured event — land in a follow-up PR in the `pocketpaw` repo.

## Files

- `src/soul_protocol/spec/decisions.py` — three Pydantic payload models (`AgentProposal`, `HumanCorrection`, `DecisionGraduation`), two builders (`build_proposal_event`, `build_correction_event`), two journal queries (`find_corrections_for`, `trace_decision_chain`), and a simple tag-co-occurrence clusterer (`cluster_correction_patterns`) that surfaces graduation candidates.
- `tests/test_spec/test_decisions.py` — 19 tests covering JSON round-trips, `Literal` enforcement on `disposition`, builder output shape, causation-vs-correlation-id filtering, chain ordering and non-decision-event filtering, and clusterer threshold / `since` behavior.
- `docs/decision-traces.md` — concept doc with the three event types, an example flow, why this matters versus raw training data, and what is deliberately out of scope for this slice.
- Re-exports on `soul_protocol.spec` and the top-level `soul_protocol` package.

## What is deliberately not here

- **Promotion logic.** The clusterer surfaces candidate patterns. Writing a `decision.graduated` event, plus choosing between `semantic` and `core` tiers, is a follow-up once the graduation policy is spec'd.
- **Embedding-based clustering.** Current match is exact on the sorted tag tuple. Richer schemes are a future slice.
- **Edit-distance scoring.** The field exists on `HumanCorrection`; no scorer is bundled. Consumers compute what they want and attach it.
- **pocketpaw wire-in.** Follow-up PR in the `pocketpaw` repo.

## Notes for review

- The `ACTION_NAMESPACES` catalog already included `agent.proposed`, `human.corrected`, and `decision.graduated` per the RFC appendix, so no change to `journal.py` was needed.
- Builders accept an optional `ts` / `event_id` so callers can pre-seed for deterministic replay; otherwise UTC-aware `datetime.now(timezone.utc)` and `uuid4()` are used.
- `find_corrections_for` pulls a large window (`limit=10_000`) and filters in-process on `causation_id` rather than pushing it into the backend query — the backend indexes action but not causation_id as a first-class column today. A real busy org will want a scoped `since` filter; callers can drop to the backend query directly if needed.
- `cluster_correction_patterns` is intentionally ~40 LOC. Not a heuristics rabbit hole.

## Base

Targets `feat/journal-engine` (PR #166). Will retarget to `dev` once that merges.

## Test plan

- [x] `uv run pytest tests/test_spec/test_decisions.py -v` — 19 / 19 pass
- [x] `uv run pytest tests/` — 1979 / 1979 pass

Refs #164 (RFC).

### DIFF
--- a/docs/decision-traces.md
+++ b/docs/decision-traces.md
@@ -1,0 +1,126 @@
+<!--
+decision-traces.md — Concept doc for the decision-trace event chain.
+Created: feat/decision-traces — Workstream D of Org Architecture RFC (PR #164).
+Pairs with src/soul_protocol/spec/decisions.py and tests/test_spec/test_decisions.py.
+-->
+
+# Decision Traces
+
+> Every agent proposal a human edits, accepts, rejects, or defers becomes a
+> structured, auditable, replayable pair of events in the org journal. Over
+> time, recurring corrections graduate into standing guidance the agent
+> loads as memory.
+
+This is one of the four compounding data types that differentiate Paw OS
+from generic stack-of-record systems. Decisions are not built in SAP or
+Salesforce. They are built here.
+
+---
+
+## The three event types
+
+A decision trace is a chain of journal events linked by `causation_id`:
+
+1. **`agent.proposed`** — payload: [`AgentProposal`](../src/soul_protocol/spec/decisions.py)
+   The agent emits a structured proposal: a tool call, a draft message, or
+   a decision among options. The proposal carries a one-to-three sentence
+   summary, the structured payload, an optional self-reported confidence,
+   the alternatives the agent considered, and references to prior journal
+   events the agent consulted (`context_refs`).
+
+2. **`human.corrected`** — payload: [`HumanCorrection`](../src/soul_protocol/spec/decisions.py)
+   A reviewer dispositions the proposal — `accepted`, `edited`, `rejected`,
+   or `deferred`. The event's `causation_id` points at the
+   `agent.proposed` event. `structured_reason_tags` (e.g.
+   `tone_too_formal`, `wrong_recipient`) carry the *why* in a form the
+   graduation pipeline can cluster on.
+
+3. **`decision.graduated`** — payload: [`DecisionGraduation`](../src/soul_protocol/spec/decisions.py)
+   When a pattern of corrections recurs, the system promotes it from
+   episodic memory (the raw correction events) to semantic or core memory
+   (standing guidance the agent reads on every relevant turn). The payload
+   carries the pattern summary, the supporting correction ids (so the
+   rule is auditable), the target tier, and the scope it applies to.
+
+---
+
+## Example flow
+
+The sales agent at Acme drafts a reply to a buyer:
+
+```text
+ts=T0  action=agent.proposed       payload=AgentProposal(kind="message_draft",
+                                                         summary="Reply to pricing question",
+                                                         proposal={...})
+ts=T1  action=human.corrected      causation_id=<T0.id>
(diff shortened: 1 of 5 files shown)


prakashUXtech (commented):
**Verdict:** LGTM with one process note and one spec concern

**Top findings**

1. **PR base targets `feat/journal-engine`, not `dev`.** This is correct and the PR description explains it: #168 stacks on #166 (journal engine) which stacks on dev. The merge order matters — if #166 merges but #168 retargets before the journal engine API stabilizes, the `find_corrections_for` / `trace_decision_chain` / `cluster_correction_patterns` helpers will break because they call `journal.query(action=..., limit=..., since=...)` with a keyword interface that must match whatever `JournalBackend.query()` exposes. The test suite uses a real SQLite journal fixture, so this is exercised, but worth confirming the `query(since=...)` kwarg is in the engine's public interface before #168 merges.

2. **`cluster_correction_patterns` pulls `limit=10_000` with no `scope` filter.** The PR description acknowledges this ("callers can drop to the backend query directly if needed"), but at scale a busy org will have far more than 10K correction events, meaning the function silently undersamples without any signal to the caller. Consider returning a `{"truncated": True}` hint in the output dict or accepting a `scope` filter argument. The current behavior is safe for v1 given the engine's expected deployment size, but naming it explicitly in the docstring as a known limitation would help consumers.

**Nits**

- `find_corrections_for` also uses `limit=10_000` without a `since` parameter. The same caveat applies; at minimum, thread `since` through to the backend call when the caller provides it. The function signature doesn't accept `since` today, which means the caller has no escape hatch short of querying the backend directly.
- The `docs/decision-traces.md` concept doc includes a line: "Decisions are not built in SAP or Salesforce. They are built here." This reads more like positioning copy than a spec doc. Fine for a blog post; in a protocol specification it may age poorly. Suggest softening to describe the structural difference (agent proposals are a first-class schema primitive in the journal, not an audit trail retrofit) without the competitive framing.
- `DecisionGraduation.supporting_correction_ids` uses `Field(min_length=1)` — good. Worth verifying this actually validates list length in Pydantic v2 (it does, via `min_length` on list fields) but the test for it (`test_decision_graduation_requires_supporting_ids`) confirms it works. No action needed.

**Positive notes**

The causation/correlation ID discipline here is exactly right. `build_correction_event` requires `causation_id` as a non-optional kwarg, making it impossible to emit a `human.corrected` event that isn't linked to a proposal. The test `test_find_corrections_for_ignores_same_correlation_id` is particularly valuable — it explicitly guards against the easy mistake of filtering on `correlation_id` instead of `causation_id` when looking up corrections. That test should survive any future refactor of the query helpers.

The `cluster_correction_patterns` function is appropriately minimal. Keeping the clustering to exact tag-tuple matching rather than introducing embedding or edit-distance logic in this slice was the right call.

PR #169, opened 2026-04-13
feat(engine): retrieval router + credential broker for Zero-Copy

prakashUXtech:
## What this adds

The retrieval router and credential broker that make the RFC's Zero-Copy federation story actually dispatch. `DataRef` already exists as a payload variant on `EventEntry` (#165), but nothing yet fans out a `retrieval.query` across local projection sources and live external sources. This PR is that plumbing.

Three modules land:

- `spec/retrieval.py` — `CandidateSource`, `RetrievalRequest`, `RetrievalCandidate`, `RetrievalResult`. Source-specific payload stays opaque; the router never interprets it.
- `engine/retrieval/router.py` — `RetrievalRouter` with `first` / `parallel` / `sequential` strategies, per-source timeout, scope overlap filtering, and a `retrieval.query` journal event on every dispatch.
- `engine/retrieval/broker.py` — `CredentialBroker` protocol plus `InMemoryCredentialBroker`. TTL-bounded credentials, scope-overlap enforcement, and `credential.acquired` / `credential.used` / `credential.expired` events when a journal is attached.
- `engine/retrieval/adapters.py` — the `SourceAdapter` protocol, a `MockAdapter` for tests, and a `ProjectionAdapter` for the local-view case (soul memory, kb, fabric).

## What's deliberately out of scope

- No concrete external adapter. Drive, Salesforce, and friends are C2 and ship separately once this plumbing is on `dev`.
- No caching layer. `RetrievalCandidate.cached` is already on the model so adapters can flag cache hits, but the cache store is its own PR.
- No async. Everything is sync for this slice; the `parallel` strategy uses `concurrent.futures.ThreadPoolExecutor`. Async wrapping is a follow-up.
- No `RetrievalTrace` integration. The trace module from #161 isn't on this base branch, so `RetrievalResult.trace` stays `Any | None = None` with a TODO.

## Tests

16 new tests under `tests/test_engine/test_retrieval.py` covering:

- all three dispatch strategies (parallel merge by score, `first` short-circuits after a non-empty reply, `sequential` stops at `limit`)
- scope filtering — a request scoped to `org:sales:leads` skips a source registered under `org:support:*`
- per-source timeout — a slow adapter is abandoned and shows up in `sources_failed`
- `retrieval.query` journal emission with the right action, scope, actor, correlation id
- broker acquire / ensure_usable / mark_used / revoke + journal emission for each
- cross-scope refusal (`CredentialScopeError`) and TTL expiry (`CredentialExpiredError`)
- Protocol conformance for both adapters
- end-to-end DataRef source flow where the router acquires and marks a credential before calling the adapter

Full suite is still green on everything I touched. The six pre-existing failures in `test_cognitive_adapters.py` and `test_mcp_sampling_engine.py` also fail on the base branch without this diff.

## Stacking

Base is `feat/journal-engine` (#166). After that merges, I'll retarget this to `dev`. This is the second PR in the Workstream C chain; the concrete Drive and Salesforce adapters will follow in C2.

RFC: #164.

### DIFF
--- a/src/soul_protocol/engine/journal/__init__.py
+++ b/src/soul_protocol/engine/journal/__init__.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 from .backend import JournalBackend
 from .exceptions import IntegrityError, JournalError, NotFoundError, SchemaError
 from .journal import Journal, open_journal
+from .scope import scope_matches, scopes_overlap
 from .sqlite import SQLiteJournalBackend
 
 __all__ = [
@@ -18,4 +19,6 @@ __all__ = [
     "SQLiteJournalBackend",
     "SchemaError",
     "open_journal",
+    "scope_matches",
+    "scopes_overlap",
 ]
--- a/src/soul_protocol/engine/journal/scope.py
+++ b/src/soul_protocol/engine/journal/scope.py
@@ -1,0 +1,69 @@
+# scope.py — Public scope-matching helper extracted from the sqlite backend.
+# Created: feat/retrieval-router — lift `_scope_matches` out of the SQLite
+# backend (where router and broker were reaching into a private symbol) and
+# into a proper public module. Semantics are unchanged; the sqlite backend
+# now imports from here, and the router + broker do too.
+#
+# Grammar (colon-delimited segments, ``*`` is a single-segment wildcard):
+#   - ``"org:sales"`` matches exactly ``org:sales``.
+#   - ``"org:sales:*"`` matches ``org:sales:leads`` and ``org:sales:deals``
+#     but NOT bare ``org:sales`` (segment count must match).
+#   - ``"org:*"`` matches ``org:sales`` and ``org:hr`` but NOT bare ``org``.
+#   - ``"org:*:leads"`` matches ``org:sales:leads`` and ``org:hr:leads``.
+#
+# This is a placeholder until ``spec.scope`` from #162 lands; the final
+# grammar there is a superset of what we implement. Callers swap to the
+# richer helper when it ships — same name, same shape.
+
+from __future__ import annotations
+
+
+def scope_matches(event_scopes: list[str], query_scopes: list[str]) -> bool:
+    """Return True iff any event scope matches any query pattern.
+
+    Semantics are asymmetric on purpose: `event_scopes` are concrete scopes
+    carried by a journal event, `query_scopes` are patterns supplied by a
+    caller looking for matches. Wildcards are valid on the pattern side;
+    putting a wildcard in an event scope works but means "this event was
+    emitted across a wildcard scope" which few writers do.
+    """
+    for pat in query_scopes:
+        pat_parts = pat.split(":")
+        for scope in event_scopes:
+            scope_parts = scope.split(":")
+            if len(pat_parts) != len(scope_parts):
+                continue
+            if all(p == "*" or p == s for p, s in zip(pat_parts, scope_parts)):
+                return True
+    return False
+
+
+def scopes_overlap(granted: list[str], requested: list[str]) -> bool:
+    """Return True iff any requested pattern is covered by any granted one.
+
+    Policy pinned: this is intentionally symmetric against wildcards — a
+    credential granted for ``org:sales:*`` is usable by a requester
+    presenting a specific scope like ``org:sales:leads``, AND a credential
+    granted for a specific scope ``org:sales:leads`` is usable by a
+    requester presenting ``org:sales:*`` (a wildcard requester asserting
+    "I am operating anywhere in this subtree").
+
+    Both directions are deliberate. Breaking either:
+      * reject specific-requester-vs-wildcard-grant -> admins who issue a
+        broad credential can't use it for any concrete call.
+      * reject wildcard-requester-vs-specific-grant -> a retrieval router
+        operating org-wide can't consume per-scope credentials, which
+        kneecaps fanout.
+    """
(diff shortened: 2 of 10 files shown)


prakashUXtech (commented):
**Verdict:** Request changes. One security-critical issue in the broker's audit path, one correctness issue in scope enforcement, and a private import that will cause breakage on refactor.

---

**Top findings**

**1. [Security-Critical] Broker silently drops audit events when the journal is unavailable**

In `InMemoryCredentialBroker._emit()`:

```python
try:
    self._journal.append(entry)
except Exception:
    pass  # Auditing should never crash the broker.

**Review by prakashUXtech — COMMENTED** [review]
**Verdict:** Request changes. One security-critical issue in the broker's audit path, one correctness issue in scope enforcement, and a private import that will cause breakage on refactor.

---

**Top findings**

**1. [Security-Critical] Broker silently drops audit events when the journal is unavailable**

`InMemoryCredentialBroker._emit()` catches all exceptions and continues. The consequence: if the journal is down or full, credential.acquired / credential.used / credential.expired events are silently lost. For a Zero-Copy system where the journal *is* the audit trail, this is fail-open. A revoked credential can be used and the use goes unrecorded.

The comment acknowledges this ("Log surface is #161's concern") but shipping it silently fail-open is a problem even for a v0 slice. The minimum acceptable fix is one of:
- If a journal is configured, a failed emit should propagate (fail-closed). Callers that want fire-and-forget should pass `journal=None`.
- Or: buffer failed emissions in-memory and expose a `flush()` / `drain_failed()` for operators. That at least makes the loss visible.

The same pattern appears in `RetrievalRouter._emit_query_event()`. Retrieval audit loss is less severe (it's a query log, not a security event), so the swallow there is defensible. The credential lifecycle events are a different category.

**2. [Correctness] Scope enforcement is unidirectional; a broader credential passes a narrower check**

`_scopes_overlap(granted, requested)` asks: does any *requested* scope satisfy any *granted* pattern? That means a credential issued for `org:*` (a wildcard) will be accepted by a requester presenting `org:sales:leads` (a specific scope), which is correct. But a credential issued for `org:sales:leads` (specific) will *not* be accepted by a requester presenting `org:*` (wildcard), because the check only runs grant-as-pattern over request-as-value.

The real problem surfaces when the requester's scope is broader than the credential's scope. A credential scoped to `org:sales:leads` should *not* be usable by a requester claiming `org:*`. But with the current check, `_scopes_overlap(["org:sales:leads"], ["org:*"])` returns False — the credential is refused, which happens to be the right outcome. Work through the inverse: `_scopes_overlap(["org:*"], ["org:sales:leads"])` returns True — a wildcard-scoped credential passes for a specific requester. Whether that is intentional is not documented. The asymmetry should have a comment with the security rationale, or be made symmetric with a documented policy on which direction grants authority.

**3. [Structural] Router imports a private function from sqlite.py**

```python
from soul_protocol.engine.journal.sqlite import _scope_matches
```

`_scope_matches` is a module-private helper (leading underscore). The router reaching into the journal's SQLite module creates a hidden coupling: renaming or extracting `_scope_matches` to its own module will silently break the router. Move `_scope_matches` to a shared location (`soul_protocol.engine.journal.scope` or `soul_protocol.spec.scope`) and export it, or duplicate it intentionally with a clear comment linking the two sites. The broker already has the correct instinct — it duplicates the function with a comment explaining why. The router should do the same rather than importing across the private boundary.

---

**Test gaps**

- No test for audit-emit failure. There should be a test where the journal raises on `append()` and asserts the broker raises (if fail-closed is the chosen policy) or records the loss (if buffered).
- No test for the specific scope asymmetry case: credential scoped `org:*`, requester presenting `org:sales:leads`. Add a test that documents the expected outcome and pins the policy.
- `revoke()` emits `credential.expired` rather than a dedicated `credential.revoked` action. The test `test_broker_emits_journal_events` checks that `credential.expired` appears after `revoke()`, which means revocations and natural TTL expiry are indistinguishable in the audit log. Either add a separate action name or document why the merge is intentional.
- No test confirming `Credential.token` does not appear in exception messages. `CredentialScopeError` and `CredentialExpiredError` messages only include `credential.id`, `source`, `scopes`, and `expires_at` — that is correct — but a test asserting `token not in str(exc)` would lock that in.

---

**Nits**

- `_with_timeout` for `first` and `sequential` strategies spawns a new `ThreadPoolExecutor(max_workers=1)` per source call. That is one `os.thread_start` per adapter invocation. For `sequential` over many sources this adds up. A single shared executor on the router, or a reusable per-dispatch one, would be cleaner.
- `parallel` strategy iterates `future_to_name.items()` which is dict insertion order in Python 3.7+, but the semantics of parallel dispatch do not depend on ordering, so this is fine — just worth a comment that the order does not affect correctness.
- `MockAdapter` stores `self._raises: Exception | None` but the type hint allows passing an instance. The common pattern is passing the class and calling it; storing an instance is fine, but the docstring should clarify callers pass an instance, not a class.

---

**Positive notes**

The three-strategy dispatch shape is well thought out. `first` short-circuits correctly, `sequential` respects `limit` at accumulation time (not just at truncation), and `parallel` uses a bounded executor sized to the source count rather than a fixed-width pool. The per-source timeout wrapping for `first` and `sequential` via `_with_timeout` is clean and daemon-safe. The `Credential.token` field is correctly kept out of all journal payloads and error messages in the current implementation. The `ProjectionAdapter` callable wrapper is exactly the right level of abstraction for local views.

PR #170, opened 2026-04-13
feat(onboarding): soul org full wizard + status/destroy + root undeletability

prakashUXtech:
## What this does

Brings `paw os init` up to the 8-step wizard from RFC #164, adds the rest of the org lifecycle commands, and wires in the three layers of root-agent undeletability.

Stacks on PR #167 (`feat/paw-os-init`). Targets `dev` once the chain merges.

### Init wizard

The minimal bootstrap kept org name, purpose, root soul, keys, and two genesis events. This PR layers on:

- **Org values** prompt (3-5, comma-separated) + `org.values_set` event
- **Founder user soul** with `--founder-name` / `--founder-email`, exported to `~/.pocketpaw/users/<name>.soul`, plus `user.joined` and `user.admin_granted` events
- **First-level scopes** (up to five, e.g. `org:sales`, `org:ops`, `org:me`) each emitting their own `scope.created`
- **Starter fleet stub** — sales / support / solo / skip — emits a placeholder `agent.spawned` with a note pointing at pocketpaw#940 for the real install
- **Invite hint** in the summary (`paw user invite <email>`); the command itself is a stub that exits non-zero so callers know it isn't wired yet

Every new step has a flag so non-interactive bootstrapping still works end-to-end.

### Status and destroy

- `paw os status [--json]` — reads the journal and reports org name, purpose, values, root DID, event count, last event ts, user count, agent count, and the scope tree
- `paw os destroy --confirm --i-mean-it` — refuses without both flags. In interactive mode you also have to type the org name. Tarballs the org dir to `~/.pocketpaw/archives/org-destroyed-<utc-stamp>.tar.gz`, then removes the original

### Root undeletability (RFC #164)

Three layers, all advisory at this stage but consistent:

1. **File** — `Identity.role` is a free-form string. Root soul births with `role="root"`. `Soul.delete(path)` peeks at `soul.json` (and `manifest.json` stats for encrypted archives) and raises `SoulProtectedError` when role matches root.
2. **Protocol** — `spec.journal.check_root_undeletable(entry, root_did)` rejects `agent.retired` and `soul.deleted` events targeting the root DID via `actor.id`, `payload.target_did`, `payload.soul_id`, or `payload.did`.
3. **CLI** — new `soul delete <path>` calls `Soul.delete()`, prints a clear error, and exits non-zero for root souls. Non-root souls delete normally with `--yes` to skip the prompt.

The tarball-and-remove path through `paw os destroy` is the only sanctioned way to take a root down.

## Test plan

- [x] `tests/test_cli/test_onboarding_full.py` — 14 new tests covering wizard, status, destroy, and all three undeletability layers
- [x] `tests/test_cli/test_paw_os_init.py` — original 8 tests still pass
- [x] Full suite green (1876 passed; the 6 pre-existing failures are missing optional deps `httpx` and `mcp`, unrelated)

## Notes

- ~944 LOC across runtime, spec, CLI, and tests. A touch over the 900 budget — happy to split status/destroy into a follow-up if reviewers want a smaller diff
- All datetimes are tz-aware UTC
- No new dependencies
- `paw user invite` is intentionally a stub (exits 1 with a message) — wiring real invites is a separate PR

## Follow-ups

- Real fleet install needs pocketpaw#940
- `paw user invite` flow (token + email + acceptance)
- Consider promoting `check_root_undeletable` from advisory helper into a journal-engine `append()` hook once we have a `root_did` known to the engine

### DIFF
--- a/src/soul_protocol/cli/main.py
+++ b/src/soul_protocol/cli/main.py
@@ -31,6 +31,7 @@ from __future__ import annotations
 import asyncio
 import io
 import json
+import sys
 import zipfile
 from datetime import datetime
 from pathlib import Path
@@ -623,6 +624,29 @@ def retire(path, preserve_memories):
         console.print(f"[yellow]{name}[/yellow] has been retired. Thank you for the memories.")
 
     asyncio.run(_retire())
+
+
+@cli.command("delete")
+@click.argument("path", type=click.Path(exists=True))
+@click.option("--yes", is_flag=True, help="Skip the confirmation prompt.")
+def delete_cmd(path, yes):
+    """Delete a .soul file from disk.
+
+    Refuses for souls with role='root' (use `soul org destroy` instead).
+    """
+    from soul_protocol.runtime.exceptions import SoulProtectedError
+    from soul_protocol.runtime.soul import Soul
+
+    if not yes and not click.confirm(f"Permanently delete {path}?", default=False):
+        console.print("[dim]Cancelled.[/dim]")
+        return
+
+    try:
+        Soul.delete(path)
+    except SoulProtectedError as exc:
+        console.print(f"[red]error:[/red] {exc}")
+        sys.exit(1)
+    console.print(f"[yellow]Deleted[/yellow] {path}")
 
 
 @cli.command()
--- a/src/soul_protocol/cli/org.py
+++ b/src/soul_protocol/cli/org.py
@@ -1,39 +1,35 @@
-# cli/org.py — `soul org init` command: bootstrap an org.
-# Renamed: feat/paw-os-init — was cli/paw_os.py. Flattened the Click group from
-#   `soul paw os <cmd>` to `soul org <cmd>` and moved the default data dir from
-#   ~/.pocketpaw/org/ to ~/.soul/. Set SOUL_DATA_DIR to override. Governance
-#   persona description updated from "Paw OS instance" to "org instance".
+# cli/org.py — `soul org {init,status,destroy}` and `soul user invite` stub.
+# Renamed: feat/paw-os-init — was cli/paw_os.py. Flattened the Click group
+#   from `soul paw os <cmd>` to `soul org <cmd>`, promoted `user` to a
+#   top-level sibling group, and moved the default data dir from
+#   ~/.pocketpaw/org/ to ~/.soul/. Set SOUL_DATA_DIR to override.
 # Created: feat/paw-os-init — Workstream A slice 3 of the Org Architecture RFC (#164).
-#
-# This command brings an org into existence from nothing:
-#   1. Create the org directory (default ~/.soul/, or $SOUL_DATA_DIR)
-#   2. Birth a root governance soul and save it as root.soul
-#   3. Generate an Ed25519 signing keypair for the root identity
-#   4. Initialize the SQLite journal via open_journal()
-#   5. Append the genesis event chain: org.created + scope.created(org:*)
-#
-# Scope is intentionally narrow — this is not the full 8-step wizard from the
-# RFC. It is the minimum viable bootstrap so follow-up slices (starter fleet,
-# user joins, policy bootstrap) have a real org to build on.
-#
-# TODO(slice #163 integration): when `load_template("governance")` ships in
-# the installed package, replace the hardcoded persona below with a template
-# load so organizations can customize their root agent prior to init.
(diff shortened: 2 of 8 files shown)


prakashUXtech (commented):
**Verdict:** Two blockers — one is a data-loss scenario in the default destroy path, one is a PII write to the append-only journal

**Summary**

The wizard extension is well-structured and the three-layer undeletability implementation is thorough. The \`check_root_undeletable\` helper covers the four payload fields where a root DID can hide, and the test suite covers all three layers including the actor-as-target path that often gets missed. Before merge, two issues need fixing.

---

**Blocker 1 — \`org destroy\` self-includes the archive it is writing**

\`_default_archives_dir()\` returns \`~/.soul/archives/\`, which is a subdirectory of \`_default_data_dir()\` (\`~/.soul/\`). \`_archive_org\` calls \`tarfile.open(archive_path, "w:gz")\` and then \`tf.add(data_dir, recursive=True)\`. With default paths, \`data_dir\` is the parent of \`archives_dir\`, so the tar run recurses into \`~/.soul/archives/\` and attempts to include the file it is currently writing. \`tarfile\` raises \`tarfile.ReadError\` at that point because the file descriptor is open for write.

The test passes only because it supplies an explicit \`--archives-dir\` outside \`data_dir\`. The default-path scenario is untested and broken.

Fix: before calling \`tf.add\`, check whether \`archives_dir.is_relative_to(data_dir)\` and either exclude the \`archives_dir\` subtree via the \`filter\` argument or move the default archives location outside \`~/.soul/\` (e.g. \`~/.soul-archives/\`).

---

**Blocker 2 — founder email written verbatim to the append-only journal**

\`user.joined\` event payload includes \`"email": founder_email or ""\`. Because the journal is append-only, there is no mechanism to remove it later. This is a GDPR right-to-erasure problem and, practically, means anyone with read access to \`journal.db\` gets the founder's email address in plaintext.

Fix: omit \`email\` from the journal payload entirely. The founder user soul (\`users/Pat.soul\`) already holds the email in its persona text, which is the appropriate place for it. If the email is needed for audit purposes, hash it (\`sha256\`) before storing.

---

**Paw-leak audit**

No new leaks introduced by this PR. The leaks that exist (\`docs/paw-os.md\`, \`docs/getting-started.md\`) were inherited from \`feat/paw-os-init\` and are called out in the #167 review.

---

**Undeletability — bypass path check**

All three layers work as designed for the normal delete path. One edge case worth noting: \`_peek_soul_role\` silently returns \`""\` on any \`OSError\`, \`BadZipFile\`, or \`JSONDecodeError\`. A deliberately corrupted \`root.soul\` would fall through and allow \`Soul.delete()\` to proceed. This is acceptable for the current "advisory" framing but should be promoted to a hard error once the protection is considered mandatory.

Layer 2 (\`check_root_undeletable\`) is not wired into \`journal.append()\` — it is a standalone helper that callers invoke manually. The docstring says "advisory" so this is in scope, but the follow-up to promote it to a journal-engine hook should be tracked.

---

**What's good**

- Test coverage for all three undeletability layers is excellent, including the actor-as-target retire path.
- \`org destroy\` requiring both \`--confirm\` and \`--i-mean-it\` plus a typed name prompt is a sensible friction design.
- \`_gather_status\` is a pure function with no side effects — easy to test and reason about.
- \`soul user invite\` exits non-zero with a clear tracking reference. Right call for a stub.
- The step-by-step journal event chain (org.created → values_set → user.joined → admin_granted → scope.created per scope → agent.spawned) gives a clean audit trail that is fully reconstructable.
- \`SOUL_DATA_DIR\` is honored consistently across \`init\`, \`status\`, and \`destroy\` via the shared \`_default_data_dir()\` helper.

---

**Nits (non-blocking)**

- \`_peek_soul_role\` appends \`role or ""\` to \`candidates\` from \`soul.json\` before checking \`manifest.json\`, then returns \`candidates[0]\` at the end. If \`soul.json\` is present but has no role and \`manifest.json\` is absent, this returns \`""\` correctly — but the \`candidates\` list is doing no real work here. A plain early-return after the manifest check would be clearer.
- Partial-state on failed init is the same as in #167: if \`journal.append\` raises after the root soul and keys are on disk, \`--force\` is the only recovery. Worth a single line in error output.
- Founder email in \`_build_founder_user_soul\` persona text (\`f"Reachable at {email}"\`) means it also lands in the soul file — that is more appropriate than the journal, but worth calling out so a future privacy pass knows where to look.

Issue #218, opened 2026-04-30
Release tracker: 0.5.0 — Intelligence + Integrity


prakashUXtech:
Tracker for the v0.5.0 minor release. Major bump because the schema grows (#192 brain-aligned memory primitives) and several new top-level subsystems land (soul-aware evals, soul optimize autoresearch, graph traversal API).

## Theme

Two pillars:

1. **Intelligence** — souls evaluate themselves (#160), improve themselves (#142), and traverse their own graph memory (#108 + #190).
2. **Integrity** — better memory updates (#192), tighter trust chain (#199-#205), audit-log polish (#201, #202), chain pruning (#203 stub), key rotation foundation (#204).

Plus quality-of-life CLI surfaces (#189 \`soul journal\`, #191 \`soul diff\`).

### Trust chain hardening + observability

| Issue | PR | Branch | Status |
|---|---|---|---|
| #199 + #200 + #205 + #204 | #210 | feat/trust-chain-verification-hardening | Open against dev |
| #201 + #202 | #213 | feat/audit-log-and-logging-improvements | Open against dev |
| #203 (stub) | #214 | feat/touchtime-chain-pruning | Open against dev |

### Memory primitives (#192)

| Step | Branch | Status |
|---|---|---|
| RFC | #209 (rfc/memory-update-primitives) | Open against dev — captain reviewing |
| Spike | feat/memory-update-primitives-spike | Pending — waits on RFC sign-off |
| Production | feat/memory-update-primitives | Pending — waits on captain daily-test of spike |

### Intelligence

| Issue | PR | Branch | Status |
|---|---|---|---|
| #160 | #211 | feat/soul-aware-evals | Open against dev |
| #108 + #190 | TBD | feat/graph-memory-traversal | In flight |
| #142 | TBD | feat/soul-optimize-autoresearch | Bench — waits on #211 merge |

### CLI polish

| Issue | PR | Branch | Status |
|---|---|---|---|
| #189 | #212 | feat/soul-journal-cli | Open against dev |
| #191 | #215 | feat/soul-diff-cli (stacked on #212) | Open against dev |

## Schema migrations

\`soul migrate <path>\` covers 0.4.x → 0.5.0:

- \`MemoryEntry.retrieval_weight: float = 1.0\` (new in #192) — backfilled from \`importance\` for legacy entries
- \`MemoryEntry.supersedes: str | None = None\` (new in #192) — derived from existing \`superseded_by\` reverse-mapping
[...]

PR #232, opened 2026-05-05
feat(test): Adding cross runtime tests suite for PocketPaw, LangChain, CrewAI

### LINKED ISSUE #228: feat(test): cross-runtime test suite — PocketPaw, LangChain, CrewAI
## Why

A portable file format only matters if it actually works across runtimes. Round-trip tests across more than one agent framework are the proof, and right now there are none.

## Scope

Add a `tests/cross_runtime/` directory with one adapter fixture per runtime. Each adapter exports a populated soul, then imports it from `.soul` into the target runtime, then asserts that shared memory items round-trip. PocketPaw integration runs at the subprocess level; LangChain and CrewAI run through their public Python APIs. Mark the suite optional under `@pytest.mark.cross_runtime` so contributors without all three runtimes installed do not see CI failures.

## Acceptance

- `tests/cross_runtime/README.md` documents the setup steps for each runtime
- PocketPaw round-trip passes for semantic and episodic memory
- LangChain and CrewAI round-trip passes for at least semantic memory (their memory models are simpler)
- CI marks the suite optional with `@pytest.mark.cross_runtime`

## Out of scope

- Feature parity across runtimes. The bar is round-trip survival, not behavioral equivalence.

## Notes for mentees

This is one of four LFDT Decentralized Trust Mentorship deliverables for soul-protocol. See `docs/lfdt-mentorship.md` (landing in PR shortly) for the full scope. Tracking: LF-Decentralized-Trust-Mentorships/mentorship-program#75.

Akshat-Raj:
## What does this PR do?

This PR adds a test suite for cross runtime integration for PocketPaw, LangChain, CrewAI.

Fixes #228

## How to test

```
uv run --python 3.12 --with pytest-asyncio pytest tests/cross_runtime/ -v
```

## Checklist

- [x] Tests pass locally (`uv run pytest tests/`)
- [x] Lint passes (`uv run ruff check . && uv run ruff format --check .`)
- [x] No secrets or credentials in the diff
- [x] PR title follows Conventional Commits (`feat:`, `fix:`, `docs:`, etc.)


prakashUXtech (changes requested):
Thanks for taking on the cross-runtime suite. The structure is right: marker registered, importorskip in place, three runtimes covered. Two things need fixing before this can merge, plus a handful of polish items.

## Blockers

**1. Retarget to `dev`, not `main`.**

All v0.5.0 work goes through `dev` per release tracker #218 ("Sub-PRs target `dev`, not main directly"). Both of your PRs target main right now. For this PR: rebase the branch onto `origin/dev` and update the PR base. (I'll make sure this is louder in the contributing docs so future mentees don't trip on the same thing.)

**2. The PocketPaw test doesn't actually test PocketPaw.**

Issue #228 says "PocketPaw integration runs at the subprocess level." `test_pocketpaw.py` currently calls `Soul.awaken` and `soul.export` through the soul-protocol library directly. That tests soul-protocol's own round-trip, not whether PocketPaw can consume the file. The test would pass even if PocketPaw was completely broken.

To make it a true cross-runtime test, the PocketPaw side should either:
- Subprocess to the `paw` CLI and verify it can load the soul, or
- Import the pocketpaw Python runtime (separate package) and verify the persona / memory loads through PocketPaw's APIs

The LangChain and CrewAI tests have the right shape: they go through the target runtime's API. PocketPaw should follow the same pattern.

## Smaller items

- **`soul.remember()` in `test_pocketpaw.py:18`** is being deprecated by #236. Switch to `soul.note(...)`.
- **Inconsistent skip pattern.** `test_langchain.py` uses `pytest.importorskip`, `test_crewai.py` uses `try/except ImportError + pytest.skip(allow_module_level=True)`. Both work, but `importorskip` is the canonical pytest pattern. Make them consistent.
- **Private API access in `test_pocketpaw.py:34`**: `recovered_soul._memory._semantic.facts()`. Tests that reach into `_underscore` members break on internal refactors. The cleaner shape is `recovered_soul.recall(query=..., type=...)` or whatever public API exists.
- **`readme.md` → `README.md`** per convention (matches what the issue spec uses).
- **Missing newline at EOF** in `readme.md`.
- **Memory round-trip assertion missing for LangChain and CrewAI.** Issue acceptance: "LangChain and CrewAI round-trip passes for at least semantic memory." Your current tests only verify persona persistence in the system prompt string. Add at least one semantic-memory assertion per runtime.

## What's working

- Marker registration in conftest.py is exactly right
- `pytest.importorskip` for the LangChain side is idiomatic
- README explains the optional-dependency story clearly so future contributors know why some tests skip locally
- The round-trip kernel test (export, re-awaken, assert) is the right shape

Once the base retarget + the PocketPaw subprocess test land, plus the small polish, this is solid. Good first pass on the cross-runtime story.

PR #237, opened 2026-05-10
feat(trust): adding digital signature verification on import

### LINKED ISSUE #227: feat(trust): digital signature verification on import
## Why

A `.soul` file should fail closed when its trust chain is broken or absent. Today verification is opt-in, which means an unverified file can quietly load and start serving memories as if it were trusted. Verification should run by default at import time and refuse to load on hard failure.

## Scope

Extend the existing trust chain code in `src/soul_protocol/runtime/trust/` so that `import` runs verification before returning a soul. Return a structured `VerificationResult` covering the passed, warned, and failed states with a reason field. Refuse to import on hard failure unless the caller passes `--allow-unverified`, in which case emit a `WARNING` log line. Reference: `docs/trust-chain.md` for the existing chain semantics.

## Acceptance

- `import` runs verification by default
- `VerificationResult` is a typed dataclass with at least `status`, `signer`, `reason`
- Hard failures (tampered chain, missing signature) refuse import with a non-zero exit
- `--allow-unverified` works and emits a `WARNING` log line
- Tests cover a clean chain, missing signature, tampered chain, and key rotation mid-chain

## Out of scope

- Designing a new chain format. The current chain code stays as-is; this issue wires verification into the import path.

## Notes for mentees

This is one of four LFDT Decentralized Trust Mentorship deliverables for soul-protocol. See `docs/lfdt-mentorship.md` (landing in PR shortly) for the full scope. Tracking: LF-Decentralized-Trust-Mentorships/mentorship-program#75.

Akshat-Raj:
## What does this PR do?

This PR shifts the trust chain verification from a "fail-open" to a "fail-closed" security posture. It wires trust chain verification directly into the Soul.awaken() import path. If a .soul file has a tampered, missing, or broken trust chain, it will now hard-fail and refuse to load by raising a SoulTrustError. This strict verification can be bypassed using the newly introduced --allow-unverified flag, which will emit a WARNING log instead.

Fixes #227 

## How to test

```
# Run the specific trust verification tests
uv run pytest tests/test_trust_chain/test_import_verification.py -v
```

## Checklist

- [x] Tests pass locally (`uv run pytest tests/`)
- [x] Lint passes (`uv run ruff check . && uv run ruff format --check .`)
- [x] No secrets or credentials in the diff
- [x] PR title follows Conventional Commits (`feat:`, `fix:`, `docs:`, etc.)

### DIFF
--- a/src/soul_protocol/cli/main.py
+++ b/src/soul_protocol/cli/main.py
@@ -3146,7 +3146,7 @@ def verify_cmd(path, as_json):
         from soul_protocol.runtime.soul import Soul
         from soul_protocol.spec.trust import chain_integrity_check
 
-        soul = await Soul.awaken(path)
+        soul = await Soul.awaken(path, allow_unverified=True)
         summary = chain_integrity_check(soul.trust_chain)
 
         # Compute time span (first → last entry)
@@ -3221,7 +3221,7 @@ def audit_cmd(path, action_prefix, limit, as_json):
     async def _audit():
         from soul_protocol.runtime.soul import Soul
 
-        soul = await Soul.awaken(path)
+        soul = await Soul.awaken(path, allow_unverified=True)
         log = soul.audit_log(action_prefix=action_prefix, limit=limit)
 
         if as_json:
--- a/src/soul_protocol/cli/main.py
+++ b/src/soul_protocol/cli/main.py
@@ -3130,7 +3130,10 @@ def repair_cmd(
 @cli.command("verify")
 @click.argument("path", type=click.Path(exists=True))
 @click.option("--json", "as_json", is_flag=True, default=False, help="Emit machine-readable JSON.")
-def verify_cmd(path, as_json):
+@click.option(
+    "--allow-unverified", is_flag=True, default=False, help="Bypass strict validation during load."
+)
+def verify_cmd(path, as_json, allow_unverified):
     """Verify the trust chain of a .soul file or directory.
 
     \b
@@ -3146,7 +3149,7 @@ def verify_cmd(path, as_json):
         from soul_protocol.runtime.soul import Soul
         from soul_protocol.spec.trust import chain_integrity_check
 
-        soul = await Soul.awaken(path, allow_unverified=True)
+        soul = await Soul.awaken(path, allow_unverified=allow_unverified)
         summary = chain_integrity_check(soul.trust_chain)
 
         # Compute time span (first → last entry)
@@ -3207,7 +3210,10 @@ def verify_cmd(path, as_json):
 )
 @click.option("--limit", type=int, default=None, help="Show only the most recent N entries.")
 @click.option("--json", "as_json", is_flag=True, default=False, help="Emit machine-readable JSON.")
-def audit_cmd(path, action_prefix, limit, as_json):
+@click.option(
+    "--allow-unverified", is_flag=True, default=False, help="Bypass strict validation during load."
+)
+def audit_cmd(path, action_prefix, limit, as_json, allow_unverified):
     """Print a human-readable timeline of signed actions.
 
     \b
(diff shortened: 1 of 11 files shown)


prakashUXtech (changes requested):
The architecture matches the spec (structured `VerificationResult`, three-state enum, `--allow-unverified` bypass with WARNING log), and the test matrix exercises all five acceptance scenarios from #227. Couple of blockers and a handful of polish items before it can land.

## Blockers

**1. Retarget to `dev`, not `main`.**

Same workflow miss as your #232. All v0.5.0 work goes through `dev` per tracker #218. Rebase + update the PR base.

**2. Legacy souls (empty chain) shouldn't be FAILED.**

`TrustChainManager.verify()` now returns `VerificationState.FAILED` when `chain.entries` is empty (`trust/manager.py`). The test `test_legacy_archive_without_chain_still_loads` was updated to expect FAILED and pass `allow_unverified=True` to make it load.

The issue says "refuse to import on hard failure", but a pre-trust-chain legacy soul isn't tampered, it's just from before the trust chain existed. Conflating "no chain" with "broken chain" means anyone with a v0.3.x soul archive has to use `--allow-unverified` forever to load it. That defeats the security posture: users train themselves to always pass the flag, then miss real tampering.

Suggested fix: map empty chain to `WARNED` instead of `FAILED`. That logs a heads-up about a legacy file but still loads. Keep `FAILED` for actually-broken chains (tampered signatures, key mismatches).

## High-priority cleanup

**3. The `VerificationResult` tuple shim deserves a deprecation path.**

`__iter__` + `__eq__` against tuples let `passed, reason = result` and `result == (True, None)` keep working for old callers. Two issues with the shim as it stands:

- No removal path. It'll be in the API forever unless someone files a follow-up issue.
- The `__eq__` against tuples is asymmetric. `result == (True, None)` works (calls `VerificationResult.__eq__` first), but `(True, None) == result` calls Python's tuple `__eq__` first, gets `NotImplemented`, and falls back. Equality should be symmetric.

Either grep for the few tuple-unpacking callers (should be under 10 sites) and migrate them now, or leave the shim with a `TODO: remove in 0.6.0 (see #XXX)` comment pointing at a follow-up issue.

## Polish

- **Branch name `cli` is too generic.** Future you (or whoever rebases on top) will thank you for `feat/issue-227-trust-import-verification` instead.
- **Thinking-out-loud comments in `test_import_verification.py:81-83`.** The "Need to update payload hash..." → "Wait..." → "So it should..." three-line train-of-thought is great while writing the test, but should be cleaned up to a clear one-line docstring before posting. Worth a self-check pass before the next push.
- **`tests/test_cli/test_org_init.py`** has an unrelated whitespace change (removed blank line between two tests, line 95). Revert to keep the diff focused on this issue.
- **`SoulTrustError.__str__` drops the signer.** The exception captures `signer` as an attribute, but the message format is `f"Trust chain validation failed: {reason}"`. Operators triaging from logs would benefit from including signer in the message too.
- **CHANGELOG entry missing.** This is both an addition (the verification + `SoulTrustError`) and a behavior change (default refuses tampered souls). Add lines under both `## Added` and `## Changed`.
- **`--allow-unverified` is only on `verify` and `audit` commands.** Other CLI entry points that load souls (`inspect`, `status`, `recall`, etc.) will hard-refuse on a broken soul with no escape hatch. Intentional? If so, doc it (verify/audit are the diagnostic paths, others enforce strict). If not, add the flag.

## What's working

- The structured `VerificationResult` dataclass + three-state enum matches the issue spec exactly
- Test matrix is the right one (clean / tampered / missing / key-rotation / bypass-warns)
- Adding `lifecycle.birth` chain entry in `birth()` and `lifecycle.reincarnate` in `reincarnate()` is a thoughtful touch: removes the edge case of "freshly-birthed has no chain"
- Honest test updates (`test_encryption.py`, `test_back_compat_load.py`, `test_persistence.py` passing `allow_unverified=True` explicitly). The dependency is visible rather than hidden.
- New `SoulTrustError` with `reason` + `signer` fields makes downstream error handling cleaner

Once the base retarget + the legacy-souls semantics + the tuple-shim path are sorted, this is mergeable.

PR #268, opened 2026-06-10
fix(cli): resolve Windows UnicodeEncodeError in soul status (#267)

### LINKED ISSUE #267: bug: `soul status` crashes on Windows with UnicodeEncodeError
## Description

`soul status` crashes on Windows due to a `UnicodeEncodeError` when the Rich library 
attempts to render emoji/special characters through the legacy Windows console.

## Environment

- **OS:** Windows 11
- **Python:** 3.13.5
- **Branch:** dev (v0.5.0.dev0)
- **Rich version:** 15.0.0

## Steps to Reproduce

```bash
soul birth "TestAgent" -o test.soul
soul status test.soul

sshekhar563:
## What

Fixes `soul status` and `soul inspect` crashing on Windows with `UnicodeEncodeError`.

Fixes #267

## Root Cause

Windows legacy console uses `cp1252` encoding. Rich library outputs Unicode characters 
(█, ░, emoji) that `cp1252` can't encode → crash.

## Fix

Reconfigure `stdout`/`stderr` to UTF-8 on Windows before Rich Console initialization 
(3 lines, zero risk to Linux/Mac).

## Testing

- `soul status test.soul` → ✅ Works (was crashing before)
- `soul inspect test.soul` → ✅ Works  
- 229/231 CLI tests pass (2 pre-existing env failures unchanged)

### DIFF
--- a/src/soul_protocol/cli/main.py
+++ b/src/soul_protocol/cli/main.py
@@ -126,6 +126,7 @@ from rich.panel import Panel
 from rich.table import Table
 from rich.text import Text
 
+
 # Fix #267: Legacy consoles (Windows cp1252, Linux C/POSIX locale, Docker,
 # cron jobs, SSH sessions) cannot render Unicode characters (progress bars,
 # emoji) that Rich outputs. Reconfigure to UTF-8 when the current encoding
--- a/src/soul_protocol/cli/main.py
+++ b/src/soul_protocol/cli/main.py
@@ -140,6 +140,7 @@ def _ensure_utf8(stream):
         except (AttributeError, OSError):
             pass  # Non-reconfigurable stream (e.g. piped output)
 
+
 _ensure_utf8(sys.stdout)
 _ensure_utf8(sys.stderr)
 
--- a/src/soul_protocol/cli/main.py
+++ b/src/soul_protocol/cli/main.py
@@ -126,6 +126,15 @@ from rich.panel import Panel
 from rich.table import Table
 from rich.text import Text
 
+# Fix #267: Windows legacy console uses cp1252 which cannot render Unicode
+# characters (progress bars, emoji) that Rich outputs. Reconfigure to UTF-8.
+if sys.platform == "win32":
+    try:
+        sys.stdout.reconfigure(encoding="utf-8")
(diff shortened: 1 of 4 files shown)


prakashUXtech (changes requested):
Thanks for reporting this and turning around a fix. Reconfiguring stdout to UTF-8 before the Rich `Console()` is the right approach, and it covers every command, not just `soul status`.

Two changes before it goes in:

**1. A regression test.** This reproduces off Windows too: running `soul status` in a subprocess with `PYTHONIOENCODING=cp1252` triggers the same crash on Linux and macOS. So a test can live in `tests/test_cli/` and actually run in CI.

**2. Gate on the encoding, not the platform.** The same `UnicodeEncodeError` fires on Linux under a C or POSIX locale (Docker images, cron jobs, ssh sessions where stdout is ASCII), which the `sys.platform == "win32"` check leaves broken. Checking `sys.stdout.encoding` instead covers both cases, and it makes the regression test above runnable on the Linux CI runners (the matrix is ubuntu-only, so a Windows-gated fix can't be exercised there at all).

Something like reconfiguring when `(stream.encoding or "").lower().replace("-", "") != "utf8"` for each of stdout and stderr.

One heads-up that isn't on you: the reason none of this showed as a failing check is that CI doesn't run tests on PRs targeting `dev` right now. I'm fixing that in a separate PR.

prakashUXtech:
This has been sitting since July, so a nudge. The two asks from the last review still stand and neither is large:

- [ ] Gate on the encoding, not the platform. Check whether `sys.stdout.encoding` can handle the characters (or just whether it is not UTF-8) instead of `sys.platform == "win32"`, so Linux under a C or POSIX locale gets the same fix.
- [ ] A regression test. Run `soul status` in a subprocess with `PYTHONIOENCODING=cp1252` and assert exit 0. That reproduces the crash on Linux and macOS, so it can live in `tests/test_cli/` and run in CI.

It also needs a rebase, dev has moved a fair bit.

If you would rather drop this one, say so and I will close it and leave #267 open for whoever picks it up next. Either answer is fine, I just do not want it to rot.

PR #274, opened 2026-06-21
fix(cli): default include_keys to False in soul export (#273)

### LINKED ISSUE #273: security: `soul export` includes private keys by default
## Bug

All CLI export commands pass `include_keys=True` to `soul.export()`, which 
includes the private Ed25519 signing key in every exported `.soul` file.

The docstring in `soul.py` warns:
> "Set to True only when you explicitly want to hand off signing power 
> (e.g. migrating between your own devices)."

## Impact

Anyone who receives an exported `.soul` file gets the private key and 
can impersonate that soul by signing new trust chain entries.

## Fix

Add `--include-keys` flag to `soul export` (default: False). 
Private keys are only included when the user explicitly opts in.

sshekhar563:
## What

Fixes #273

The `soul export` command was passing `include_keys=True` by default, which included the private Ed25519 signing key in every exported `.soul` file.

## Changes

- Added `--include-keys` flag to `soul export` (default: `False`)
- Private keys are only included when user explicitly passes `--include-keys`
- Help text warns: "Only use when migrating between your own devices"

## Usage

```bash
# Safe default — no private keys in exported file
soul export aria.soul

# Explicit opt-in for device migration
soul export aria.soul --include-keys


prakashUXtech (changes requested):
Good catch on this one. Exporting the private signing key by default was a real leak, and the fix is right. I went through all the `include_keys=True` call sites on `dev` to check nothing else was affected, and the flip is complete: `soul export` was the only share-intended path that leaked the key, and the in-place saves (`remember`, `evolve`, and the rest) correctly keep it. `--include-keys` still works for migration. Clean change.

Two things before merge, both packaging rather than substance:

**1. A test.** Since this is a security default, it's worth locking in. A CliRunner test asserting the default export zip contains `keys/public.key` but not `keys/private.key`, and that `--include-keys` puts the private key back. `tests/test_cli/` is the spot.

**2. Docs.** `docs/cli-reference.md` under `soul export` doesn't list `--include-keys`, and the "Portable transfer between machines" line now needs the flag to carry signing power. Worth updating both here.

One suggestion, non-blocking: the migrate-then-delete workflow now silently drops signing power, and with the quiet verification-only mode it's easy to miss until signing stops working. A one-line console notice on default export ("private key not included; pass `--include-keys` to migrate signing power") would save someone a confusing afternoon.

Also a stray extra blank line snuck in around `main.py:705`. Drop it while you're in there.

Thanks for the threat write-up on the issue too. Made this easy to verify.

prakashUXtech (approved):
Verified: the test asserts private.key is absent by default and present with --include-keys, and flipping the default turns exactly that test red — so it's a real guard. The rebase is clean (dev is a full ancestor, and the --password/_awaken_or_fail work from #294 plus the deprecation warning all survive), and the console notice is in and tested.

This fix was verified complete two rounds ago; it only ever needed the test and the doc. Sorry it sat as long as it did — that was on my gate, not your code. Merging.

One doc nit for a follow-up, not blocking: --include-keys is in the usage example and the format-table wording is fixed, but it's still missing from the Options table in cli-reference.

PR #276, opened 2026-06-27
feat(cli): add `soul did show` and `soul did verify` commands

sshekhar563:
## What

Phase 1 of the DID CLI feature — adds `soul did` command group for inspecting 
and verifying a soul's decentralized identity.

### `soul did show <source>`
Displays the soul's identity info:
- Name, DID (`did:soul:...`), archetype, born date
- Ed25519 public key (truncated)
- Private key status (present/absent)
- Trust chain entry count and key rotation status

### `soul did verify <source>`
Verifies the soul's DID integrity:
- DID format validation (`did:soul:` prefix)
- Public key presence check
- Trust chain signature verification via `Soul.verify_chain()`
- `--verbose` flag shows individual chain entries

## Tests

15 tests added — all passing:
- `TestDidShow` (7 tests): name, DID, algorithm, public key, private key status
- `TestDidVerify` (5 tests): valid soul, no-keys, verbose, error handling
- `TestDidGroup` (3 tests): help text for all commands

## Files Changed

- **[NEW]** `src/soul_protocol/cli/did.py` — DID CLI module
- **[MODIFY]** `src/soul_protocol/cli/main.py` — Register `did` command group
- **[NEW]** `tests/test_did_cli.py` — 15 test cases

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -312,6 +312,57 @@ soul export aria.soul --output aria.md --format md
 
 ---
 
+### `soul did show`
+
+Display the DID (Decentralized Identifier) and identity info of a soul. Shows name, DID, public key, algorithm, private key status, key rotation history, and trust chain length.
+
+```bash
+# Show DID identity for a .soul file
+soul did show aria.soul
+
+# Show DID identity for an unpacked soul directory
+soul did show .soul/
+```
+
+**Arguments:**
+
+| Argument | Required | Description |
+|----------|----------|-------------|
+| `SOURCE` | Yes | Source soul file (`.soul`) or unpacked directory. |
+
+---
+
+### `soul did verify`
+
+Verify the trust chain and DID integrity of a soul. Checks that:
+- The DID format is valid (`did:soul:...`)
+- A public key is present
+- All trust chain entries have valid Ed25519 signatures and the hash chain is intact
+
+Exits with code 1 if any check fails.
+
+```bash
+# Verify a soul's DID and trust chain
+soul did verify aria.soul
+
+# Verify with detailed chain table
+soul did verify aria.soul --verbose
+```
+
+**Arguments:**
+
+| Argument | Required | Description |
+|----------|----------|-------------|
+| `SOURCE` | Yes | Source soul file (`.soul`) or unpacked directory. |
+
+**Options:**
+
+| Option | Short | Description |
+|--------|-------|-------------|
+| `--verbose` | `-v` | Show a detailed table of trust chain entries with sequence numbers, actions, timestamps, and signatures. |
+
+---
+
 ### `soul migrate`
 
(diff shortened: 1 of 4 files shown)


prakashUXtech (changes requested):
Nice work getting `soul did` off the ground. The Rich output panels read well, and the small conventions are all in place: lazy `Soul` import inside the command, the file-header comment, `click.Path(exists=True)` matching the other commands.

There are a couple of correctness issues in `did verify` that I hit by running the branch, so I'd like those sorted before it goes in.

**1. `did verify` exits 0 even when verification fails.** Pointed at a soul with a tampered trust chain, it prints the red "Verification failed" panel but still returns exit code 0. The existing `soul verify` returns 1 in the same case. Anything scripting `soul did verify x.soul && ...` would treat a forged soul as valid, so this needs `sys.exit(1)` on any failed check.

**2. `did verify --verbose` crashes on a soul that has chain entries.** The verbose table reads `entry.ts`, but the field on `TrustEntry` is `timestamp`, so it raises `AttributeError` on the first non-empty chain.

**3. The test fixture's trust chain is empty, which is why 1 and 2 slipped past 15 green tests.** The fixture builds the soul with `birth()` + `remember()`, and `remember()` doesn't append trust-chain entries. So every test runs the empty-chain path and `verify_chain()` never actually executes. Using `observe()` in the fixture gives you a real chain to verify against, plus a tampered-chain case to assert the non-zero exit.

Smaller things:

- ruff fails on the branch: unused `Path` and `Text` imports, and `import sys` placed after `console = Console()` trips E402. Moving the `sys` import to the top clears it.
- Docs: `docs/cli-reference.md` and `CHANGELOG.md` should pick up the new commands in this PR.
- `soul verify` already verifies chains with exit codes and `--json`, and `soul audit` already renders the chain timeline. The real new value here is the DID and key display. Worth considering whether `did verify` should delegate the chain check to that shared path (you'd get exit codes and `--json` for free), or a note in the PR on why you'd rather keep it separate.

One heads-up that isn't on you: none of the ruff issues showed up as a check because CI doesn't currently run lint or tests on PRs targeting `dev`. That's a repo gap, and I'm fixing it in a separate PR.

prakashUXtech:
Same nudge as #268, this has been untouched since July. The blockers from the last review:

- [ ] `soul did verify` must exit 1 when verification fails, matching `soul verify`. Right now a tampered chain prints the red panel and still exits 0, which breaks anything scripting `soul did verify x.soul && ...`.
- [ ] The other correctness items from that review (see above in the thread).
- [ ] Tests for both commands, including the failing-verification exit code.
- [ ] Rebase on dev.

If the DID CLI is no longer something you want to carry, tell me and I will close this and file the remaining scope as an issue so the design is not lost. Finish or close, either is fine.

PR #295, opened 2026-07-04
Fix/graph recall rebind 281

sshekhar563:
## Description
This PR resolves Issue #281 where graph-augmented recall silently degraded after a Soul was restored from its state.

### The Bug
During initialization (`MemoryManager.__init__`), the `RecallEngine` was instantiated with a reference to an empty `KnowledgeGraph`. 
When `MemoryManager.from_dict()` restored a state, it completely rebuilt the graph and rebounded `manager._graph`, but the `manager._recall_engine` continued holding onto the old, empty graph. Because of this stale reference, graph-augmented recall returned empty results for restored souls, even while new observations mutated the new graph correctly.

### The Fix
- Rewired the graph reference inside `from_dict()` immediately after the `KnowledgeGraph` is restored: `manager._recall_engine._graph = manager._graph`.
- Confirmed that this is the lowest-risk approach for deserialization (other built-in stores like episodic/semantic are updated in-place via `.model_validate` so they do not suffer from this rebinding hazard).

### Testing Added
- Added a regression test `test_from_dict_rewires_recall_engine_graph` to `tests/test_memory/test_memory.py`.
- The test explicitly mutates a graph, serializes it, and asserts that `restored._recall_engine._graph is restored._graph` after calling `from_dict`.

Fixes #281

### DIFF
--- a/src/soul_protocol/runtime/memory/manager.py
+++ b/src/soul_protocol/runtime/memory/manager.py
@@ -1,4 +1,8 @@
 # memory/manager.py — MemoryManager facade orchestrating all memory subsystems.
+# Updated: 2026-07-11 (#281) — Rewire recall engine graph reference after
+#   from_dict(). The RecallEngine held a stale reference to the pre-restore
+#   empty KnowledgeGraph, so graph-augmented recall returned nothing after
+#   awaken/reincarnate.
 # Updated: 2026-06-16 (feat/soul-skills-procedural) — adds
 #   consolidate_agent_procedures(): a curator pass over PROCEDURAL memories that
 #   only ever considers AGENT-authored entries (provenance == AGENT). It finds
@@ -2409,8 +2413,8 @@ class MemoryManager:
         graph_data = data.get("graph", {})
         if graph_data:
             manager._graph = KnowledgeGraph.from_dict(graph_data)
-            # Re-wire the recall engine to point to the restored graph
-            manager._recall_engine._graph = manager._graph
+            # Re-wire the recall engine to point to the restored graph (#281)
+            manager._recall_engine.rebind_graph(manager._graph)
 
         self_model_data = data.get("self_model", {})
         if self_model_data:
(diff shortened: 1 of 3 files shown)


prakashUXtech (changes requested):
The fix is correct, and it closes the issue for both `awaken` and `reincarnate`. Pointing `_recall_engine._graph` at the same restored graph object is the right move: the engine reads the graph live at query time, so recall now sees the restored entities, and because you share the object rather than copy it, later `observe()` calls that mutate the graph in place stay visible too. I checked the other rebind sites (`__init__`, `clear`) and the engine is the only long-lived holder of the reference, so fixing `from_dict` is enough. Good diagnosis.

One thing I'd like before merge: the test asserts reference identity (`_recall_engine._graph is _graph`) but never calls `recall()`, so it doesn't actually exercise the symptom the issue is about, which is graph-augmented recall going silent after a restore. If someone later refactors the engine to read its graph from a different attribute, this fix line would set a dead attribute, recall would break silently, and the test would still pass. Could you add a behavioral assertion: after `from_dict`, seed a memory that mentions a graph-connected entity, call `recall(...)`, and assert the graph-augmented result comes back? That's the one that guards the real regression.

Two conventions while you're in there: add an `Updated: … (#281)` line to the file headers, and give the PR a Conventional Commits title (something like `fix(memory): rewire recall engine graph after awaken (#281)`), since it becomes the squash commit subject.

Optional and non-blocking: reaching into the engine's private `_graph` works fine, but a small `RecallEngine.rebind_graph()` (or rebuilding the engine the way `clear()` does) would be a touch cleaner. Your call.

prakashUXtech (approved):
This is exactly what I was hoping to see. The behavioral test now seeds an `Alice -> Bob` graph edge, stores a Bob memory, and after the round-trip calls `recall("Alice")` and asserts the Bob memory surfaces through graph augmentation. That fails without the rewire (Alice never matches the Bob content directly, so it can only come back via the graph), so it genuinely guards the silent-degradation symptom rather than reference identity. Headers are updated, and you went the extra step with a public `RecallEngine.rebind_graph()` instead of poking the private attribute. Nicely handled.

Approving. One tiny thing for the squash: the PR title is still "Fix/graph recall rebind 281" and it becomes the commit subject, so I'll set it to `fix(memory): rewire recall engine graph after awaken (#281)` when I merge. Thanks for the quick turnaround.

Issue #42, opened 2026-03-06
feat: Trust chain — verifiable action history with signatures


prakashUXtech:
## Summary

A soul should be able to prove what it learned and where. Trust chain adds cryptographic verification to a soul's history — every learning event, memory mutation, and evolution step is signed and traceable.

## Spec Addition (spec/)

New protocol primitives:

- TrustEntry — signed record of an action (memory write, personality change, learning event)
- TrustChain — append-only log of TrustEntry items, Merkle-tree verifiable
- SignatureProvider protocol — pluggable signing backends (Ed25519, DID-based, etc.)
- Verification functions — verify_chain(), verify_entry(), chain_integrity_check()

```
trust_chain/
  entry_001.json  # { action, timestamp, actor_did, signature, prev_hash }
  entry_002.json
  ...
```

## Runtime Addition (runtime/)

- Ed25519SignatureProvider — default signing backend
- TrustChainManager — appends entries on memory writes, evolution events, learning events
- CLI: soul verify <path> — verify chain integrity of a .soul file
- CLI: soul audit <path> — human-readable history of what changed and when

## Why This Matters

- Grant-winning feature for NLnet — "verifiable AI agent cognition"
- Enables provenance: "this soul learned X from source Y on date Z, signed by runtime W"
- Foundation for reputation systems — other agents can verify trust before interacting
- Differentiator: nobody else combines portable identity + verifiable history
- Aligns with DID-based identity already in spec

## Dependencies

- Builds on Learning Events (issue above) — learning events are a key thing to sign
- Uses existing crypto/ module as foundation
- Pairs with EternalStorage — archived souls carry their trust chain

## Sizing: XL

prakashUXtech:
Update post-v0.3: the append-only signed-history half is shipped.

Now in dev:
- Journal hash-chain, each event carries prev_hash linking to the prior event
- Root Agent governance posture (no conversational surface, signed governance events only, see docs/org-journal-spec.md)
- ACTION_NAMESPACES catalog covering org.created, key.rotated, and friends
- EventEntry has an optional sig field already; onboarding generates keys for the root agent

What's still open:
- SignatureProvider protocol (separable from the hash-chain that already exists)
- Default Ed25519 backend wired to actually sign per-event
- soul verify and soul audit CLI commands
- Making sig required rather than optional once the backend is in place

Re-scope: "implement the signing layer on top of the existing hash-chain." The chain is the hard part and it's done; signing is the follow-up.

[...]

Issue #41, opened 2026-03-06
feat: User-defined memory layers + domain isolation


prakashUXtech:
## Summary

The protocol should not dictate memory structure. Layers should be user-defined namespaces, not a hardcoded enum. Domains should isolate context within layers.

Right now MemoryType is a fixed enum: CORE, EPISODIC, SEMANTIC, PROCEDURAL. The brainstorm vision says the SDK should give you soul.layer("procedural").store(...) where layer names are arbitrary.

## Spec Changes (spec/)

- Replace fixed MemoryType with open string-based layer names
- Add domain field to MemoryEntry (optional namespace like "finance", "legal", "personal")
- MemoryStore protocol accepts layer + domain as query parameters
- .soul file format: memory/{layer_name}/{domain}/ directory structure

## Runtime Changes (runtime/)

- Default layers: core, episodic, semantic, procedural (backward compatible)
- New social layer for relationship memory (bonds, trust, interaction history)
- Domain-aware MemoryManager — queries can scope to a domain
- DomainIsolationMiddleware — enforces that agents only access their authorized domains

## Why This Matters

- Makes the protocol truly general-purpose, not tied to our 4-layer opinion
- Domain isolation is Cognee's insight — agents in billing shouldn't see legal context
- Social layer enables multi-agent relationship memory
- Aligns with "HTTP not nginx" — the spec defines the namespace mechanism, not the names

## Migration

- MemoryType enum stays in runtime/ as a convenience, but spec/ uses plain strings
- Existing .soul files still work (default domain = "default")

## Sizing: L

prakashUXtech:
Update post-v0.3: the "domain isolation" half is effectively shipped via journal scope tags. Every event in the org journal carries a scope list (org:sales:*, org:engineering:foo, etc.), and recall is filtered by scope before it reaches an agent. Cognee's insight lands, just through scope rather than a separate domain field.

What still maps to this issue:
- Open string-based layer names on MemoryEntry (currently a closed MemoryType enum)
- soul.layer("custom_name").store() API on the runtime
- Default social layer for relationship-tied facts
- .soul file layout for memory/{layer_name}/ directories

Re-scoping suggestion: drop the "domain isolation" framing (scope handles it) and keep this issue as "open MemoryType + named layers." Worth deciding whether scope tags cover the original billing-vs-legal use case entirely before building MemoryType open-ness on top.

prakashUXtech:
Tracked for 0.4.0 in #183 (identity bundle).

PR #297, opened 2026-07-07
fix(memory): include archives tier in export and flat saves (fixes #283)

sshekhar563:
## Description

Fixes #283 — Export and default saves silently drop the archives memory tier.

### The Bug

`pack_soul`'s tier list, `unpack_soul`'s tier list, and the flat on-disk layout (`_FLAT_TIERS`) all omitted `archives` (consolidated conversation summaries). Even though `MemoryManager.to_dict()` emits it and `from_dict()` would restore it, the data was never written to the `.soul` archive or the flat directory layout.

Exporting a soul and re-awakening it therefore loses all archival memory. Because source entries are already marked `archived=True`, the re-archival filter skips them — the lost archives **cannot be regenerated**.

### The Fix

One-line addition in each of three locations:

| File | Change |
|------|--------|
| `src/soul_protocol/runtime/export/pack.py:120` | Added `"archives"` to the pack tier list |
| `src/soul_protocol/runtime/export/unpack.py:110` | Added `"archives"` to the unpack tier list |
| `src/soul_protocol/runtime/storage/file.py:163` | Added `("archives", [])` to `_FLAT_TIERS` |

### Tests Added

- `test_export_awaken_preserves_archives` — seeds an archive, exports to `.soul`, awakens, verifies archive survived
- `test_save_load_preserves_archives` — saves to directory, verifies `archives.json` exists on disk, loads back and checks content

### DIFF
--- a/src/soul_protocol/runtime/export/pack.py
+++ b/src/soul_protocol/runtime/export/pack.py
@@ -117,6 +117,7 @@ async def pack_soul(
                 "graph",
                 "self_model",
                 "general_events",
+                "archives",  # v0.5.0 (#283) — consolidated memory summaries
             ]:
                 default = {} if tier_name in ("graph", "self_model") else []
                 tier_data = memory_data.get(tier_name, default)
--- a/src/soul_protocol/runtime/export/pack.py
+++ b/src/soul_protocol/runtime/export/pack.py
@@ -1,4 +1,6 @@
 # export/pack.py — Create .soul zip archives from a SoulConfig.
+# Updated: 2026-07-12 (#283) — Added "archives" to the pack tier list so
+#   consolidated conversation summaries survive export round-trips.
 # Updated: 2026-04-29 (#42) — Trust chain is written to ``trust_chain/chain.json``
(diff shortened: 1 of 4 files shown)


prakashUXtech (approved):
This is a clean fix, and it's complete. I checked all three legs the issue named and each is covered: pack writes `memory/archives.json`, unpack reads it back, and `_FLAT_TIERS` now includes it so default-domain flat saves write it too. The round-trip is symmetric, the `archives` key name stays consistent from `to_dict` through `from_dict`, and older `.soul` files that have no archives file still load fine because both read paths guard on existence. The two new tests assert the archive content survives export then awaken, and flat save then load, not just that a key is present, which is exactly what I'd want to see.

CI here is green apart from one unrelated trust-chain test (`test_payload_is_hashed_not_stored`). That's a known pre-existing flake, fixed separately on dev, so you can ignore it on this PR.

One small thing before merge: the repo keeps an `Updated: <date> (#NN)` line in each file's top comment (pack.py and file.py already track #41 and #42). Add a `(#283)` line to `pack.py`, `unpack.py`, and `storage/file.py` to match the convention.

Nice work, and thanks for turning this around quickly.

PR #299, opened 2026-07-10
fix(api): add public API for CLI/MCP, stop mutating private internals (fixes #284)

sshekhar563:
## Description

Fixes #284 — CLI and MCP layers mutate private runtime internals, so refactors break them silently.

### The Bug

- CLI `main.py` had **~45** private `_`-prefixed attribute accesses (`soul._memory`, `mm._episodic`, `mm._graph._entities.clear()`, etc.)
- MCP `server.py` had **~12** similar accesses
- `repair --rebuild-graph` cleared `_entities` and `_edges` but **NOT** `_provenance` → stale provenance after rebuild
- Any internal rename in `runtime/memory` silently breaks CLI/MCP with `AttributeError` and no test catches it

### The Fix

**New public methods on `MemoryManager`:**

| Method | Replaces |
|---|---|
| `episodic_entries()` | `mm._episodic.entries()` |
| `semantic_facts()` | `mm._semantic.facts()` |
| `procedural_entries()` | `mm._procedural.entries()` |
| `social_entries()` | `mm._social.entries()` |
| `remove_episodic/semantic/procedural(id)` | `mm._episodic.remove(id)` |
| `graph_entities()` | `mm._graph.entities()` |
| `graph_remove_entity(name)` | `mm._graph.remove_entity(node)` |
| `clear_graph()` | `mm._graph._entities.clear()` (+ fixes provenance) |
| `rebuild_graph()` | 25-line inline rebuild in CLI |

**New public methods on `Soul`:**

| Method | Replaces |
|---|---|
| `memory` (property) | `soul._memory` |
| `eval_history` (property) | `soul.evaluator._history` |
| `clear_eval_history()` | `soul.evaluator._history.clear()` |
| `reset_energy()` | `soul._state.current.energy = 100.0` |
| `reset_bond()` | `soul._identity.bond.bond_strength = 50.0` |

All **~57** private accesses in CLI and MCP replaced with these public methods.

### Provenance Bug Fix

`clear_graph()` now clears `_provenance` alongside `_entities` and `_edges`. Previously, `--rebuild-graph` left stale provenance entries pointing to pre-rebuild memory IDs.

### Tests Added

- `TestPublicTierAccess` — 8 tests covering all new MemoryManager methods
- `TestSoulPublicAPI` — 4 tests covering Soul convenience methods
- `test_clear_graph_clears_provenance` — regression test for the provenance bug

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -198,6 +198,26 @@ Parse a SOUL.md-formatted string into a Soul. Does not accept an engine or searc
 | `pending_mutations` | `list[Mutation]` | Unapproved evolution proposals. |
 | `evolution_history` | `list[Mutation]` | All resolved mutations (approved and rejected). |
 
+#### MemoryManager Public API
+
+The `soul.memory` property exposes read-only access to individual memory tiers, so that interface layers don't need to reach into private stores:
+
+| Method | Returns | Description |
+|--------|---------|-------------|
+| `episodic_entries()` | `list[MemoryEntry]` | All episodic memory entries. |
+| `semantic_facts(*, include_superseded=False)` | `list[MemoryEntry]` | All semantic facts. |
+| `procedural_entries()` | `list[MemoryEntry]` | All procedural memory entries. |
+| `social_entries()` | `list[MemoryEntry]` | All social memory entries. |
+| `custom_layer_entries()` | `list[MemoryEntry]` | All entries from custom (user-defined) layers. |
+| `graph_entities()` | `list[dict]` | All entities from the knowledge graph. |
+| `graph_remove_entity(name)` | `None` | Remove a single entity from the knowledge graph. |
+| `clear_graph()` | `None` | Clear all entities, edges, and provenance from the knowledge graph. |
+| `rebuild_graph()` | `dict` | Clear and rebuild the knowledge graph from episodic + semantic memories. Returns dict with `old_count` and `new_count`. |
+| `remove_episodic(memory_id)` | `bool` | Remove an episodic memory by ID. |
+| `remove_semantic(memory_id)` | `bool` | Remove a semantic fact by ID. |
+| `remove_procedural(memory_id)` | `bool` | Remove a procedural memory by ID. |
+
+
 ### Memory Operations
 
 #### `soul.remember()`
--- a/src/soul_protocol/cli/inject.py
+++ b/src/soul_protocol/cli/inject.py
@@ -78,7 +78,7 @@ async def build_context_block(
     human = core.human or "(empty)"
 
     # Recent memories (episodic, most recent first)
-    entries = soul._memory._episodic.entries()[:memory_limit]
+    entries = soul.memory.episodic_entries()[:memory_limit]
 
     # Build memory lines
     if entries:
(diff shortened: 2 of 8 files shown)


prakashUXtech (changes requested):
This is a strong fix, and it does what #284 asked. I checked it against the base to be sure: `cli/main.py` went from 51 private-internal accesses to 0, and `mcp/server.py` from 17 to 0. Both named files are completely clean now. The `repair` path is the best part: `clear_graph()` clears `_entities`, `_edges`, and the `_provenance` dict (that last one was the stale-provenance bug), and `test_clear_graph_clears_provenance` is a genuine regression test for it. Good work.

Two corrections to the summary, both small.

1. The "last private access in the CLI" isn't quite the last one. `cli/inject.py:81` (same CLI package) still does `soul._memory._episodic.entries()[:limit]`. The public method you added covers it (`soul.memory.episodic_entries()`), it just wasn't applied there. Worth finishing the surface.

2. Three of the four highlighted methods have no callers anywhere in `src/`: `graph_edges()`, `graph_entity_count()`, and `core_memory_dict()`. Only `custom_layer_entries()` is wired in. Either connect them to the code that needs them or drop them, so we don't ship untested speculative API. (`core_memory_dict()` itself is safe, for the record: it exposes only persona and human text, no key material.)

One more that closes the issue's own stated gap: add a `CliRunner` test for `soul repair --rebuild-graph`. The unit tests protect the public methods, but a bug in the command wiring itself (a wrong method name or arg) would still ship green, and #284 specifically called out that command's missing coverage.

Smaller notes: `Soul.eval_history` returns the live `_history` list rather than a copy, so a caller could mutate internal state; return `list(...)` if it's meant to be read-only. And the file-header `Updated:` lines plus `docs/api-reference.md` should pick up the new public methods.

The core here is genuinely done and safe. These are the finishing touches.

prakashUXtech (changes requested):
Everything on the engineering side is done, and done well. `cli/inject.py` is switched to the public `episodic_entries()`, so the CLI and MCP are now genuinely free of private-internal access (I checked: zero across `main.py`, `inject.py`, and `server.py`). The three unused methods are gone, `eval_history` returns a copy, and the new `CliRunner` test for `soul repair --rebuild-graph` actually invokes the command, so it would catch a wiring mistake the unit tests can't. That is the full ask from last round.

Two hygiene items left before merge, both mechanical:

1. Add an `Updated: … (#284)` line to the top comment of the five changed source files (`cli/inject.py`, `cli/main.py`, `mcp/server.py`, `runtime/memory/manager.py`, `runtime/soul.py`).
2. In `docs/api-reference.md` you documented the new MemoryManager methods, but the four Soul-level ones (`eval_history`, `clear_eval_history`, `reset_energy`, `reset_bond`) are not in there yet.

Both are quick, and the substance is solid, so this is nearly ready.

prakashUXtech (approved):
Both hygiene items are done. The `(#284)` header line is on all five changed files, and the four Soul-level methods (`eval_history`, `clear_eval_history`, `reset_energy`, `reset_bond`) are now in `api-reference.md`. I re-checked the important invariant after the push and it holds: zero private-internal accesses across `main.py`, `inject.py`, and `server.py`, and the tier getters all return `list()` copies. No regression.

Approving. One cosmetic doc nit you can fold in whenever: the `api-reference` entry for `reset_bond` shows `def reset_bond(self) -> None` but the real signature has an optional `strength: float = 50.0`. Not worth holding for. Clean work.

Issue #203, opened 2026-04-29
trust: chain pruning + archival for long-running souls


prakashUXtech:
Found in PR #198 review. The trust chain has no size cap. \`audit_log()\` returns all entries. For long-running souls (months / years of activity), the chain grows unboundedly — large soul archives, slow verification, opaque audits.

## Fix

Add chain archival:

- New \`trust_chain/archive/\` directory with chunked archive files (e.g. \`entries_0000-0999.json\`)
- \`TrustChainManager.archive_old(before_seq: int) -> int\` moves entries to the archive and replaces them in the active chain with a single \`chain.archived\` checkpoint entry pointing at the archive file's hash
- Verification crosses the checkpoint by recomputing the archive's hash and matching the checkpoint's \`prev_hash\`
- \`soul archive-chain --before-seq N\` CLI command (parallel to \`soul archive\` for memory)
- \`soul verify\` runs across active + archive layers

## Tests to add

- Build a chain of 1000 entries, archive entries [0, 500), verify the chain is still cryptographically valid via the checkpoint linkage
- Archive integrity is preserved (manually tamper an archive file → verify fails)
- New entries can be appended after archival

## Why nice-to-have, not must-fix

Most souls won't hit problematic chain sizes for a long time. Defer until a concrete pain signal — but worth tracking so we don't forget.

Target: 0.4.x.

Issue #201, opened 2026-04-29
trust: payload summaries in audit_log for human-readable history


prakashUXtech:
Found in PR #198 review. \`Soul.audit_log()\` and \`soul audit\` CLI return entries with \`payload_hash\` (64-char hex). Useful for verification but the audit table is a wall of opaque hashes — not scannable by humans.

## Fix

Add per-action summary extraction in \`TrustChainManager.audit_log()\`. The chain manager doesn't have the original payloads (intentional — only hashes are stored), but the **caller-side append** does. Two options:

### Option A — pass summary at append time

Extend \`TrustChainManager.append(action, payload, *, summary=None)\` to accept an optional human-readable summary. Store it as a non-cryptographic field on TrustEntry (e.g. \`summary: str = ''\`). Audit log shows the summary; verification ignores it (excluded from \`compute_entry_hash\` like \`signature\`).

### Option B — derive summary from action + payload at append time

Hardcode a registry of action → summary-formatter functions:
- \`memory.write\` → \`f\"{count} memories\"\`
- \`memory.forget\` → \`f\"deleted {tier}/{id}\"\`
- \`bond.strengthen\` → \`f\"+{delta:.1f} for {user_id or 'default'}\"\`
- \`evolution.proposed\` → \`trait or proposal name\`

Apply at append time, store on TrustEntry as a non-cryptographic field.

Option B is less flexible but requires no caller changes. Recommend B.

## Tests to add

- For each action type in the repo, append one entry, verify the summary string is sensible
- Audit log displays summaries
- \`soul audit\` CLI table includes a Summary column

Target: 0.4.1.

Issue #192, opened 2026-04-27
Memory updates: brain-aligned primitive set (supersede + confirm + update + weight-decay forget)


prakashUXtech:
## TL;DR

Add a runtime-layer memory-update story modelled on how brains actually handle updates: three verbs gated by prediction error (`confirm` / `update` / `supersede`), and a forget split into retrieval-weight decay vs hard purge. Targets 0.5.x. RFC needed first.

## Why

Today, the only paths that touch the `superseded_by` field run inside the soul: `dream._dedup_semantic` during consolidation, and `manager._resolve_fact_conflicts` during a `learn`-time contradiction. Users have no API or CLI to mark a stale fact as superseded by a new one. When a held belief turns out wrong, the choices are: write a corrective memory and hope it ranks above the stale one (current workaround), or delete the old one outright and lose provenance.

The cognitive-neuroscience literature points at a more honest model. Seven mechanisms worth borrowing:

- **Reconsolidation** (Nader, Schafe & LeDoux 2000, *Nature*) — retrieval briefly destabilizes the touched memory, opening a write window of a few hours. Outside that window, the trace is stable again.
- **Prediction-error gating** (Sevenster, Beckers & Kindt 2014, *Learning & Memory*) — zero PE leaves the trace alone; moderate PE opens reconsolidation; high PE bypasses it and writes a new orthogonal trace.
- **Schema fast-tracking** (Tse et al. 2007, *Science*) — schema-consistent facts consolidate in hours, schema-conflicting facts force structural change or rejection.
- **Source monitoring** (Johnson, Hashtroudi & Lindsay 1993, *Psych. Bull.*) — old beliefs are not erased. They are tagged as "what I once thought" with a forward link to the correction. The Continued Influence Effect (Brashier et al. 2021, *PNAS*) shows belief regression tracks correction-memory retrieval strength, not the correction's presence.
- **Pattern completion vs separation** (Marr 1971; Bakker et al. 2008, *Science*) — above similarity threshold, new input merges into the existing trace. Below threshold, a new orthogonal trace is filed alongside.
- **Forgetting as inhibition** (Bjork & Bjork 1992; Wimber et al. 2015, *Nat. Neurosci.*) — "forget" is a retrieval-strength drop. Storage strength is preserved. Spontaneous recovery and reinstatement prove the trace is not gone.
[...]

Issue #191, opened 2026-04-26
feat(cli): `soul diff` — structured comparison between two souls


prakashUXtech:
## Problem

Soul files are ZIP archives. `git diff` shows them as "Binary files differ" with no insight into what actually changed. The dir-format option (`soul create --format dir`) gives line-diffable JSON, but raw JSON line-diff isn't the right granularity for soul review either. The reviewer ends up reading object literals when they want to read content.

What's missing: a `soul diff` command that compares two souls and reports what changed at the soul level. Memories added, removed, or edited. Identity changes. OCEAN drift. Bond changes. Skills. Not file-level diff, soul-level diff.

## Why this matters

Two real cases on the captain's machine right now.

### 1. Team-soul PR review (paw-workspace)

A team soul lives in git. Promoting a memory rewrites the whole file. The reviewer needs to see "what new memory is being added, and is the wording sanitised correctly." With ZIP-format souls git shows nothing. With dir-format souls git shows raw JSON. Better, but the reviewer is reading object literals when they want to read prose.

### 2. Migration verification

Re-cutting a team soul (sanitising new captain memories, propagating a corrected fact) needs a way to verify "this change set is what I intended." `soul diff old.soul new.soul` makes that possible without unpacking both archives by hand.

## Proposed surface

```
soul diff <old.soul> <new.soul> [--format text|json|markdown]
```

### Output shape (text default)

```
Identity:
  (no change)

OCEAN:
  (no change)

Core memory:
  persona unchanged
  human   unchanged

Memories:
  + semantic id=34e53f01 imp=8: "Second test memory: this is the kind of content..."
  - semantic id=ab12cd34 imp=6: "Stale fact about kb-go scopes that got superseded..."
  ~ semantic id=78ef90ab: importance 7 -> 9
  ~ procedural id=11223344: content edited (12 chars added)

Bond:
  (no change)

Skills:
  + analyst (level 1, 0 XP)

Summary: 1 added, 1 removed, 2 edited across memories; 1 skill added.
```

### `--format json`

Same data, JSON-shaped, so CI checks and PR review bots can parse it.

### `--format markdown`

Same data, rendered as a markdown table. Paste it into the PR body so reviewers see the changes at a glance.

## Git integration (follow-up)

A `.gitattributes` entry that registers `soul diff` as the diff driver for `.soul` files:

```
*.soul diff=soul
```

Combined with one git config line per developer:

```
git config diff.soul.command "soul diff"
```

[...]

PR #301, opened 2026-07-14
fix(#286): correct eternal storage CLI examples and catch unknown-tie…

sshekhar563:
Resolves #286

## Description

The eternal storage CLI examples in README.md and RFC-005 contained three bugs that caused errors when copy-pasted by new users.

### Bugs Fixed

1. **`--source` → `--tier`**: `soul recover` uses `--tier`, not `--source`. It also takes a positional reference hash (not a `.soul` path) and requires `--output`.
2. **`--tiers local,ipfs` → `-t ipfs -t arweave`**: Click's `multiple=True` requires repeated flags, not comma-separated values. Additionally, no `local` provider is registered — only `ipfs`, `arweave`, and `blockchain` exist.
3. **Unhandled `ValueError` traceback**: When `archive` is called with an unknown tier, `EternalStorageManager.archive()` raises `ValueError`. The CLI now catches this and prints a clean error message instead of a raw traceback.

### Changes

| File | Change |
|------|--------|
| `README.md` | Fixed eternal storage CLI examples |
| `rfc/RFC-005-ETERNAL-STORAGE-PROVIDER.md` | Same fix |
| `src/soul_protocol/cli/main.py` | Added `try/except ValueError` in `archive` command |
| `tests/test_eternal/test_cli_eternal.py` | Added `test_archive_unknown_tier` (clean error, no traceback) |

## Verification

- All 5 eternal storage CLI tests pass (`test_cli_eternal.py`)
- Ruff format and lint clean on all changed files


prakashUXtech (changes requested):
Good, focused fix. The README and RFC-005 examples now match the real flags (`archive -t ipfs -t arweave`, `recover <ref> --tier --output`), and the unknown-tier case is caught with a clean message that lists the available tiers instead of a traceback. Removing `local` from the docs rather than half-wiring it was the right minimal call, and it is consistent everywhere.

One thing to finish it: the same broken example still lives in `docs/cli-reference.md:1304`:

```
soul archive .soul/ --tiers ipfs arweave
```

That is the exact bug from the issue (with `multiple=True`, Click takes only `ipfs` and errors on `arweave`), and cli-reference is the doc people copy from most. One line: `soul archive .soul/ -t ipfs -t arweave`.

Optional, non-blocking: the archive-failed path prints the error but returns exit 0, so a script would not detect the failure. It matches the existing `recover` behavior, so it is not a regression, but a `raise SystemExit(1)` on both would make failures scriptable. And an `(#286)` Updated header line on `main.py` and the new test file, per convention.

Tight PR otherwise.

prakashUXtech (changes requested):
The blocker is fixed: the broken `soul archive .soul/ --tiers ipfs arweave` example in cli-reference is corrected to the repeated-flag form, and I swept the whole tree (docs, README, rfc, scripts) to confirm no broken `--tiers a b` or `--source` example survives anywhere. The optional asks are done too: the archive-failed path now does `raise SystemExit(1)` so failures are scriptable, and the `(#286)` header lines are on both files.

One thing to fix before merge, introduced by this push rather than the original work: adding the `(#286)` header line to `cli/main.py` also deleted eight earlier `# Updated:` entries (#231, #192, #142, #191, #203, #201, #189, #160), about 50 lines of changelog history. I diffed against dev to be sure. It's non-functional and CI can't see it, but it erases other people's entries, so please restore those eight and just prepend your `#286` line above them. Everything from `#42` down is still intact.

Non-blocking: a one-line test asserting the documented `-t ipfs -t arweave` two-flag form works would round it out, since the new test only covers the single-flag and failure cases. Otherwise this is clean.

prakashUXtech (approved):
Verified all three, and first: **your count was right and mine was wrong.** I said 8 deleted entries; you restored 10. The two I missed (#284 and #247) had landed on dev from the #299/#249 merges shortly before your push, so your merge resolution dropped those as well. You did the more careful accounting.

The restore checks out exactly: `main.py` at your head, minus your two new lines, diffs byte-identical against dev. Nothing lost, nothing reworded, nothing duplicated, and the commit is 62 additions with zero deletions. 27 entries where dev has 26, so the #286 line is genuinely prepended rather than replacing anything.

`test_archive_two_tier_flags` is a real test, not a shape check. The two mock providers return distinct tier names and distinct reference formats with no overlap, so a last-flag-wins regression would fail your assertion. And the original fix survives intact: I swept docs, README, rfc and scripts for any remaining `--tiers <a> <b>` or `soul recover --source` form and there are none.

Approving. Three small things you can fold in whenever, none worth holding this for:

1. `README.md:390` still lists `local` as an archive destination ("local, IPFS, Arweave, blockchain"), two lines below the block you just fixed. `local` is never registered as a provider, and your own test uses `-t local` as the *unknown* tier. So someone reading the corrected example and then trying `local` hits the exact copy-paste failure #286 exists to eliminate. One word.
2. The test file header still says only "Added test_archive_unknown_tier" but the push added a second test.
3. To fully harden both tier tests, add `assert "blockchain" not in result.output.lower()`. Without it, a regression that ignored the flags entirely would fall through to "archive to all providers" and still pass. Applies to `test_archive_specific_tier` too, which had the same pre-existing gap.

One for a separate issue, not this PR: `manager.py` validates tiers inside the archive loop, so `-t ipfs -t local` archives to IPFS first and then raises, leaving the manifest unwritten and the archive orphaned. Harmless with in-memory mocks, but a real paid Arweave provider makes that a cost leak. Worth validating all tiers up front.

Nice, careful work on the restore.

PR #308, opened 2026-08-02
fix: include social tier in deletion paths (#291)

### LINKED ISSUE #291: forget and forget_entity skip the social memory tier
Memory deletion never touches the social tier, so relationship and person memories survive a delete that is documented as covering all tiers.

`_find_entry_by_id`, `forget(query)`, `forget_entity`, `forget_before`, and `purge_by_id` all loop over episodic/semantic/procedural only; `Soul._forget_bulk` (`soul.py:2675-2689`) has the same three-tier loop. A memory stored via `remember(type=MemoryType.SOCIAL)` routes to `_social` (`manager.py:941`) and is then unreachable by any deletion path. `confirm`/`update`/`supersede`/`purge` on a social id also silently return found=False.

**Where** (commit `2ad0eb9`): `src/soul_protocol/runtime/memory/manager.py:1638`, `:1508-1510`, `:1558-1560`, `:1605-1607`, `:1878-1893`.

**Impact:** `forget_entity("Alice")` removes every episodic/semantic/procedural mention but leaves the social-tier entries, which round-trip through save/export and keep surfacing via `recall(layer="social")`. The docstrings claim all-tier / GDPR-compliant deletion, so an operator believes the data is gone when it is not. Social is the tier meant for person and relationship memories, so it is the most sensitive one to miss.

**Fix:** include the social store in `_find_entry_by_id` and every delete/purge path (or iterate the tier registry rather than a hardcoded three-tier list).

sshekhar563:
## Problem
`forget` and `forget_entity` skip the social memory tier (#291).
All deletion paths iterate a hardcoded episodic → semantic → procedural
triad, ignoring social memories entirely.

## Root Cause
When SocialStore was added in PR #41, the layer dispatch system
(_store_in_layer, _delete_in_layer, etc.) and serialization were wired
correctly, but 11 deletion/mutation functions in MemoryManager plus
Soul._forget_bulk() were never updated to include the social tier.

## Fix
1. Introduced `MemoryManager._all_stores()` — a single helper returning
   all four built-in stores as `(name, store)` tuples.
2. Rewired `forget()`, `forget_entity()`, `forget_before()`, `remove()`,
   `_find_entry_by_id()`, `forget_by_id()`, `purge_by_id()`, `count()`,
   `update_in_place()`, `set_retrieval_weight()`, and `supersede()` to
   use the helper instead of hardcoded triad iteration.
3. Updated `Soul._forget_bulk()` (v0.5.0 weight-decay path) to include
   social.

## Files Changed
- `src/soul_protocol/runtime/memory/manager.py` — primary fix + _all_stores() helper
- `src/soul_protocol/runtime/soul.py` — fix _forget_bulk() weight-decay path
- `tests/test_gdpr_deletion.py` — 10 new tests + 2 updated assertions

## Testing
- 67 tests pass (all related suites)
- New tests verify: `_find_entry_by_id` finds social, `forget`/`forget_entity`/`forget_before`/`purge_by_id` all operate on social memories, round-trip persistence (save → awaken → forget → re-awaken), and audit trail coverage.
- No regressions to existing deletion or memory tests.

## Risk
**Low.** The fix is additive. The `_all_stores()` helper is 4 lines. No public
API changes. No `RecallEngine` changes.

Closes #291

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,10 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 - **Cross-runtime test suite — PocketPaw, LangChain, CrewAI (#228, #271)** — integration tests that verify soul-protocol works correctly when embedded in PocketPaw, LangChain, and CrewAI runtimes.
 
+### Fixed
+
+- **Social tier included in deletion paths (#291)** — `forget()`, `forget_entity()`, `forget_before()`, `purge_by_id()`, `remove()`, and `count()` now include the social memory tier. Previously these functions iterated only episodic → semantic → procedural, silently skipping social memories added via `SocialStore` (#41). The return dict from bulk `forget()` and `forget_entity()` gains a `social` key (additive — existing `episodic` / `semantic` / `procedural` / `total` keys are unchanged). `soul status`, `soul forget`, `soul health` CLI commands and the `soul_forget` / `soul_health` MCP tools now report social counts. `_memory_lookup_sync` (provenance walker) and `supersede --type` also updated. A `_all_stores()` helper centralises tier iteration so future tiers are automatically included.
+
 ---
 
 ## [0.4.0] -- 2026-04-29
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -493,7 +493,7 @@ async def forget(self, query_or_id: str, *, user_id: str | None = None) -> dict
 
 **Dispatch:**
 - If `query_or_id` resolves to a known memory id, this is the single-id verb. Returns `{found, id, action: "forgotten", weight, tier}` and appends a `memory.forget` chain entry with `{id, tier, weight_after}`.
-- Otherwise it is the bulk query path (back-compat for pre-0.5.0 `Soul.forget(query)`). Every match has its retrieval_weight dropped to 0.05. Returns the legacy shape `{episodic, semantic, procedural, total}`.
+- Otherwise it is the bulk query path (back-compat for pre-0.5.0 `Soul.forget(query)`). Every match has its retrieval_weight dropped to 0.05. Returns the legacy shape `{episodic, semantic, procedural, social, total}`.
 
 The action name on the chain is unchanged from 0.4.0 (`memory.forget`); only the payload shape grows. `user_id` is recorded on the chain entry when provided.
 
@@ -531,7 +531,7 @@ Audited single-id deletion. Records a deletion audit entry when the entry exists
 
 | Key | Type | Description |
 |-----|------|-------------|
-| `episodic` / `semantic` / `procedural` | `list[str]` | The deleted ID (length 0 or 1) by tier. |
+| `episodic` / `semantic` / `procedural` / `social` | `list[str]` | The deleted ID (length 0 or 1) by tier. |
 | `total` | `int` | 0 if not found, 1 if deleted. |
 | `found` | `bool` | Whether `memory_id` resolved. |
 | `tier` | `str \| None` | Tier the entry lived in, or `None`. |
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -1000,7 +1000,7 @@ soul forget .soul/ --id bf0ee3453983 --apply              # surgical single-id
 
 Exactly one of `QUERY`, `--id`, `--entity`, or `--before` is required. Without `--apply` you get a count + per-tier breakdown without changes. With `--apply`, the runtime confirms (unless `--confirm`), writes a `.soul.bak`, deletes, saves, and reports.
 
-**Output:** A preview line (or post-action line) of the form `would forget N memories from <name>` (or `Forgot N memories ...`), followed by per-tier counts. Per-tier display reads from the result dict's `episodic` / `semantic` / `procedural` lists; if you script the runtime API directly the same shape is returned.
+**Output:** A preview line (or post-action line) of the form `would forget N memories from <name>` (or `Forgot N memories ...`), followed by per-tier counts. Per-tier display reads from the result dict's `episodic` / `semantic` / `procedural` / `social` lists; if you script the runtime API directly the same shape is returned.
 
 ---
 
@@ -1031,7 +1031,7 @@ soul supersede aria.soul "User now prefers light mode" \
 | `--reason TEXT` | Why the old memory is wrong or out-of-date (recorded in the supersede audit). |
 | `--importance / -i INT` | Importance score for the new memory (1-10, default: 5). |
 | `--emotion / -e TEXT` | Emotion tag for the new memory. |
-| `--type / -t {episodic,semantic,procedural}` | Tier for the new memory. Defaults to the old entry's tier. |
+| `--type / -t {episodic,semantic,procedural,social}` | Tier for the new memory. Defaults to the old entry's tier. |
 
 **Output:** A panel showing the old ID, new ID, tier, reason, and the new content. Saves the soul automatically. Exits non-zero if `--old-id` does not resolve.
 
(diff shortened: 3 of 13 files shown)


abhinab18 (approved):

prakashUXtech (changes requested):
This is the strongest fail-without-fix evidence in the batch — 10 of your new tests fail on dev with correct root causes, including a real save → awaken → forget → re-awaken round-trip that asserts on persisted retrieval weight. And `_all_stores()` is the right shape: one list, twelve call sites, so a fifth tier is picked up everywhere automatically. Tracing it back to when SocialStore was added was good diagnosis.

Three things before merge, and the first is structural.

**1. The branch is mis-stacked.** `fix/issue-291-social-deletion` was cut from `docs/drift-290`, not from `dev`, but declares `base: dev`. Commit `a62eff1` in this PR is byte-identical to #307's head — so merging #308 would land #290's documentation under the #291 title and leave #307 with nothing. Rebase onto `origin/dev` and drop `a62eff1`. (This is the third PR in this batch stacked on an unmerged sibling without declaring it — worth making a habit of `git log --oneline origin/dev..HEAD` before opening.)

**2. The operator-facing surfaces still under-report social — the same false-assurance the issue is about, moved up a layer.** `soul forget --entity Alice` now genuinely *deletes* social entries, but `_tier_counts` in `cli/main.py` omits social from the printed breakdown, so the operator sees nothing in social and concludes nothing was there. Same in the MCP `forget` tool's `tiers`, and forgetting a social id displays `total: 1` with all three tier lists empty. If the point is that operators shouldn't wrongly believe data is gone, they equally shouldn't wrongly believe it was never there.

**3. `Soul._memory_lookup_sync` still skips social** — and its docstring says "Synchronous lookup across all built-in stores". That's the exact bug #291 names, in a function whose docstring now actively lies. It's what the provenance walker uses to resolve a cursor, so a chain walk silently stops at a social entry.

Two regressions the PR introduces:

**`soul status` Total no longer adds up.** `count()` now includes social, and it feeds `memory_count`, but the status panel prints only three tier rows. So you get Episodic 4 / Semantic 7 / Procedural 2 and then "Total 16" with five social memories invisible. Same shape in `demo.py`. `count()` wasn't in #291's scope — either add the Social row or revert that change.

**`count()` re-arms the bug inside the helper meant to prevent it.** It iterates `_all_stores()` and then throws the iteration away:

```python
for _name, store in self._all_stores():
    if _name == "episodic":
        total += len(self._episodic._memories)
    elif _name == "semantic":
        ...
```

`store` is bound and never used, so a fifth tier added to `_all_stores()` falls through every branch and counts as zero, silently. `SocialStore.count()` exists — `total += store.count()` is what the surrounding comment already tells you to do.

Then: a CHANGELOG entry (this changes the `forget*` return shape — additive, but declare it); the `soul.py` header comment; `v0.6.0` in two comments should be `v0.5.0`; four stale doc lines in `api-reference.md` and `cli-reference.md` that still list the three-tier result shape and `supersede --type` choices; and tests for `update`/`supersede` on a social id, since the issue named those and only `purge` is covered.

One for a follow-up issue: `_all_stores()` covers the four built-ins but not custom layers or the archival store, and `BondRegistry` still holds Alice's bond record after `forget_entity("Alice")` — strength, interaction count, bonded-at, persisted to the `.soul`. Until those close, "GDPR-complete" is overclaiming; worth a `.. note::` in the docstring naming what isn't covered.

The core fix is genuinely correct — these are mechanical.

prakashUXtech (changes requested):
Good pass — the mis-stacking is gone (the diff is now three commits, all #291), and most of the list is genuinely fixed. Confirmed: `_memory_lookup_sync` covers social with a test; all three operator surfaces now report it (`_tier_counts`, the `forget --id` reshape, and the MCP tool's `tiers`), plus `supersede --type` gained social as a bonus; `count()` is now a clean `sum(store.count() for ...)` so a fifth tier can't silently count zero; there are tests for `update`/`supersede` on a social id; and the CHANGELOG entry landed in the right section. I also mutation-tested the core: removing `("social", self._social)` from `_all_stores()` turns 9 tests red, so the fix is genuinely pinned.

One correction to my own last review: the tier-table-with-Total panel is in `soul inspect`, not `soul status`. You fixed `inspect` correctly — my pointer was wrong.

Five small things left, and two of them are new:

**New — MCP `soul_health` is now internally inconsistent.** `mcp/server.py:1564` still computes `total = len(episodic) + len(semantic) + len(procedural)` while the tiers dict at `:1612` now reports social. A consumer summing the tiers won't get `total`. That's the same bug class you just fixed in `inspect`, re-created one file over.

**New — the doc fix outran the code.** `docs/cli-reference.md:1610` now says `soul health` shows "episodic, semantic, procedural, social", but `health_cmd` still reads and prints three tiers. That's drift in the opposite direction, which is a bit ironic in this PR. I'd fix the code side rather than walk the doc back, since the MCP tool already reports social.

Then: `demo.py:349-360` still lists three tier rows against a `Total` that now includes social; `soul.py`'s header comment has no `#291` entry; and `v0.6.0` should be `v0.5.0` in two comments (`manager.py:1336`, `soul.py:2715`) since the package is `0.5.0.dev0`.

Two more worth knowing. Your `forget_entity()` reshape moved `edges_removed` to the front of the return dict — harmless for dict access, but it'll shift any golden JSON snapshot, and it isn't in the CHANGELOG. And health analytics on both surfaces still build `all_mems` from three tiers, so duplicate/low-importance/orphan detection never sees social entries — pre-existing rather than yours, but the same tier-blindness and worth a follow-up issue.

Cheap addition: nothing currently pins `soul.memory_count` including a social entry, which is exactly what regressed. One assertion would cover it.

prakashUXtech (changes requested):
Most of the list is done: `soul.py`'s header carries the `#291` entry, the `v0.6.0` comments are corrected to `v0.5.0`, and MCP `soul_health` now includes social in its totals so the tier dict and `total` agree. Good pass.

Two of the items are still open, and they're the pair that undercuts the PR's own premise:

**`docs/cli-reference.md:1610` still claims `soul health` shows "episodic, semantic, procedural, social"**, but `health_cmd` still builds `all_mems = episodic + semantic + procedural`. So the doc promises a tier the code doesn't read. That's drift in the opposite direction from the one this PR exists to fix, which is why I'd rather see the code side change than the doc walked back — the MCP tool already reports social, so the CLI is the odd one out. Note this also means health's duplicate, low-importance and orphan-node analytics never see social entries at all.

**`demo.py:357-360` still lists three tier rows against `Total = soul.memory_count`**, and `memory_count` now includes social. So the demo prints rows that don't add up to its own total — the same mismatch you fixed in `soul inspect`.

Everything else checks out, and the core is well pinned: removing `("social", self._social)` from `_all_stores()` turns 9 tests red, so the deletion fix is genuinely covered.

One cheap addition while you're there: nothing currently asserts that `soul.memory_count` counts a social entry, which is exactly what regressed and got fixed. One assertion pins it.

prakashUXtech (approved):
Verified merge-ready. Both blockers from last round are closed, and you fixed the doc/code mismatch in the better direction: rather than walking the cli-reference claim back, you made `health_cmd` genuinely social-aware, so social entries now flow into the duplicate, low-importance and orphan-node analyses — not just the printed row. MCP `soul_health`'s total now equals the sum of its four tiers, and demo.py's arithmetic closes.

The mutation test is what sells it: removing `("social", self._social)` from `_all_stores()` turns 10 tests red, so the coverage is real rather than decorative. Prior nits all closed too (soul.py #291 header, v0.6.0 → v0.5.0, and the memory_count-includes-social test).

Three follow-ups worth their own issue, none blocking: MCP health analytics still build `all_mems` from three tiers so its duplicate/low-importance/orphan numbers diverge from the CLI's; that same line reaches into `_social` privately instead of the public `social_entries()` from #284; and `Soul._forget_bulk` still enumerates the four tiers explicitly rather than iterating `_all_stores()`, which is the one path that would silently miss a fifth tier.

PR #309, opened 2026-08-03
fix: exclude entity-derived skills from public_profile() and A2A cards (#292)

### LINKED ISSUE #292: public_profile() and A2A agent cards expose entity-derived names
`observe()` auto-creates a Skill from every extracted entity name (`soul.py:2105-2115`), with no type filter. `public_profile()` (documented as a leak-free third-party surface: "skill names are public, memories and relationships are not") then returns those names, and `soul_to_agent_card` (`runtime/bridges/a2a.py:51`, reachable via `soul export-a2a`, `cli/main.py:1790`) publishes them to any registry or peer.

So telling a soul "my manager Alice at Acme Corp is difficult" produces `Skill("Alice")` and `Skill("Acme Corp")`, and both names then appear in the public profile and agent card. The safety-guardrail prompt covers conversational leaks but not this structural path.

**Where** (commit `2ad0eb9`): `src/soul_protocol/runtime/soul.py:1207` (public_profile), `:2105-2115` (skill creation), `runtime/bridges/a2a.py:51`.

**Impact:** private person and organization names leak through the surface documented as safe to expose. The `public_profile` tests do not cover entity-derived skill names.

**Fix:** add a visibility flag on skills (or exclude entity-derived skills from the public surface), and separate "capabilities" from "entities extracted from memories" so the public profile only exposes intended skills.

sshekhar563:
## Problem
`observe()` auto-creates a `Skill` from every extracted entity name (`soul.py:2121-2131`), with no type filter. `public_profile()` (documented as a leak-free third-party surface) then returns those names, and `soul export-a2a` publishes them to any registry or peer.

So telling a soul *"my manager Alice at Acme Corp is difficult"* produces `Skill("Alice")` and `Skill("Acme Corp")`, and both names appear in the public profile and agent card.

Closes #292.

## Root Cause
When entity extraction runs during `observe()`, every entity creates a `Skill` with no source annotation. `public_profile()` and `A2AAgentCardBridge.soul_to_agent_card()` iterate `self._skills.skills` without filtering — they dump everything.

## Fix
1. **Added `SkillSource` enum** (`MANUAL`, `ENTITY`, `PROCEDURE`, `LEARNING`) to `skills.py`
2. **Added `source` field** to `Skill` model (defaults to `MANUAL` for backward compat)
3. **Tagged all auto-creation sites** with appropriate source:
   - `observe()` entity skills → `ENTITY`
   - `evaluate()` domain skills → `LEARNING`
   - `grant_xp_for_procedure_use()` → `PROCEDURE`
   - `grant_xp_from_learning()` → `LEARNING`
4. **Added `SkillRegistry.public_skills()`** — filters out `ENTITY` source
5. **Filtered `public_profile()`** to use `public_skills()`
6. **Filtered A2A bridge** to use `public_skills()`

## Tests Added
- `test_public_profile_excludes_entity_derived_skills` — entity skills don't leak through `public_profile()`
- `test_a2a_agent_card_excludes_entity_derived_skills` — entity skills don't leak through A2A cards
- `test_observe_tags_entity_skills_as_entity_source` — `observe()` correctly tags auto-created skills

## Verification
- All 12 tests pass in `test_safety_and_public_profile.py`
- Ruff lint: All checks passed
- Backward compatible: `source` defaults to `MANUAL`, so existing `.soul` files load without breaking


prakashUXtech (changes requested):
The design here is right and the surface audit is complete — I checked every other consumer of the skill registry (CLI `skills`/`health`, MCP tools, `diff.py`, `serialize()`, the TavernAI exporter) and you correctly left the owner-local ones unfiltered while closing both public surfaces. Tagging at all four auto-creation sites and centralising the filter in `public_skills()` is the shape I'd have picked.

The blocker is the half that decides whether this works in the real world.

**Every soul that already exists on disk still leaks.** I built a soul with dev's code, exported it, and awakened it with yours:

```
[dev ] public_profile skills: ['Acme', 'Alice', 'Corp', 'challenging_difficult']
[#309] restored:              [('alice','Alice','manual'), ('acme','Acme','manual'), ...]
[#309] public_profile skills: ['Acme', 'Alice', 'Corp', 'challenging_difficult']
[#309] A2A card skills:       ['Alice', 'Acme', 'Corp', 'challenging_difficult']
```

The restore path does `Skill(**skill_data)`, legacy entries have no `source` key, Pydantic backfills `MANUAL`, and they sail straight through the filter. The PR body frames "`source` defaults to `MANUAL`" as backward compatibility, but for a privacy fix that default is fail-open — it protects only souls born after the merge, which is the minority of the reported impact.

There's a precise, non-heuristic migration available: entity-derived skill ids are exactly `entity_name.lower().replace(" ", "_")` of graph entities, and the graph is persisted alongside the skills. So the restore path can back-fill `source=ENTITY` for any legacy skill whose id matches a known graph entity — deterministic, not a guess. Alternatives: make it `source: SkillSource | None = None` and treat unknown as non-public, or add a `soul upgrade` step since that command already exists.

**Second blocker, smaller:** `SkillRegistry.add()` is a no-op on id collision, so an ENTITY skill permanently squats the id and a later legitimate registration is silently dropped:

```
soul.skills.add(Skill(id="python", name="Python"))   # explicit, MANUAL
→ [('python','Python','entity')]                      # still entity, still hidden
```

`templates.py` registers template skills via `add()`, so a templated soul that had chatted about Python would never publish Python as a capability. Fix: upgrade `source` when an explicit MANUAL add collides with an ENTITY skill.

**On the tests** — they pass, but they don't test the reported bug. All three fail on dev only with `ImportError: cannot import name 'SkillSource'`, i.e. because the enum is absent, not because the leak reproduces. Tests 1 and 2 hand-construct `Skill(..., source=ENTITY)`, so they unit-test the filter rather than the path. There's no test that runs `observe("my manager Alice at Acme Corp")` and asserts "Alice" is absent from `public_profile()`. I confirmed the heuristic extractor fires without an API key, so that test is cheap and deterministic.

Also `test_observe_tags_entity_skills_as_entity_source` is malformed: the guard `if sk.id not in ("negotiation","empathy")` makes the `or sk.id in (...)` clause unreachable, so the assert reduces to `source != MANUAL` — which accepts `LEARNING`. A regression that tagged entity skills LEARNING would re-open the leak (LEARNING isn't filtered) and this test would stay green. Assert `source == ENTITY`.

One residual for a follow-up issue, not this PR: LEARNING-source skills stay public and their names are conversation-derived. "My HIV diagnosis came back positive" produces a public skill `diagnosis_positive`; "interviewing at Stripe, nervous" produces `interviewing_nervous`. Same class of leak, same A2A card, and your new docstring now asserts the surface is safe. Pre-existing on dev so not yours to fix here, but worth filing — and arguably LEARNING should be filtered too until domains come from a fixed taxonomy.

Then the usual: `soul.py` and `bridges/a2a.py` headers, a CHANGELOG entry, and `public_profile()`'s docstring should say entity-derived skills are excluded.

Good work — this is closing the second half, not reworking the first.

prakashUXtech (changes requested):
The P0 is closed, and you solved it better than I suggested. Rather than the graph-backfill I proposed, making `source` default to `None` and turning `public_skills()` into an allowlist is cleaner — legacy skills restore as `None` and get filtered with no id-matching and therefore no false-positive class at all. I verified with the same probe as last round: the `legacy.soul` that returned `['Acme','Alice','Corp']` on dev now returns `[]` through both `public_profile()` and the A2A card. The `add()` collision upgrade fixes the id-squat for the real path too, since templates pass MANUAL.

One thing needs to be written down before this merges, because it's a silent behaviour change for every existing soul.

A templated soul exported on dev — with genuine MANUAL-intent skills like `Python` and `Negotiation`, which have no `source` key on disk — awakens on this branch with `public_profile: []` and an **A2A card advertising zero capabilities**. That's the correct trade for a privacy fix (erring private is right), and `add()` with MANUAL upgrades in place so re-registration is the migration path. But the diff touches no CHANGELOG, docs, spec or RFC, so today it ships as a surprise: souls quietly stop advertising what they can do.

Please add a CHANGELOG entry with a migration note covering both halves: existing souls' skills become non-public until re-registered, and SDK consumers calling `add()` without an explicit `source` now get a non-public skill. Plus the top-of-file headers on `soul.py`, `bridges/a2a.py` and `templates.py`.

Non-blocking suggestions: the legacy test only asserts on `public_profile()` — extending it to the A2A card and a real export/awaken round-trip would pin both surfaces. There's a dead `if/elif/else: pass` block and an `acme_corp` id that never matches (the extractor splits it into `acme` + `corp`). And `PROCEDURE`-source skills are still public, so `grant_xp_for_procedure_use("deploy_acme_prod")` shows up in the public profile — pre-existing on dev, not yours to fix here, but worth a follow-up issue since it's the same class as #292.

prakashUXtech (approved):
Both asks are done, and the CHANGELOG entry is better than I asked for. It covers all three audiences: the behaviour change itself, the migration path for existing souls (with the exact `add(..., source=SkillSource.MANUAL)` call that upgrades in place), and the SDK-facing note that `add()` without an explicit source now creates a non-public skill. That last one is the easiest to forget and the one most likely to bite an integrator. Headers landed on all three files too.

So the P0 is closed with a clean fail-closed mechanism, the trade-off is written down, and existing users have a documented way back. Approving.

Non-blocking follow-ups, whenever you're next in here: the `PROCEDURE`-source visibility question (a skill created by `grant_xp_for_procedure_use("deploy_acme_prod")` still publishes) is worth its own issue — same class as #292, pre-existing, but now that the machinery exists it's a small change. And the dead `if/elif/else: pass` block plus the `acme_corp` id that never matches (the extractor splits it into `acme` + `corp`) are cosmetic leftovers in the test.

PR #310, opened 2026-08-04
fix: verify content-address hash on eternal recover() (#293)

### LINKED ISSUE #293: Eternal recover() does not verify content against the reference hash
Recovery from eternal storage trusts the provider to return authentic bytes.

`recover()` does `data = await provider.retrieve(source.reference); return data` with no check that `hash(data) == source.reference` for content-addressed tiers (IPFS/Arweave). The provider protocol's `verify()` only checks existence, and the CLI `recover` command writes the bytes straight to disk with no awaken/signature/trust-chain validation.

**Where** (commit `2ad0eb9`): `src/soul_protocol/runtime/eternal/manager.py:107-109`; CLI at `cli/main.py:1037`.

**Impact:** currently latent, because only mock/local providers ship (MockIPFS serves from its own in-memory dict). But the contract silently omits content-address verification, so any real gateway or pinning service wired in later inherits a substitution hole: a compromised gateway can return a different soul blob for a given CID and it is accepted. The whole point of content addressing is defeated without the check.

**Fix:** verify the retrieved bytes hash to the content-addressed reference before returning, for any content-addressed tier. Add it now so a real provider cannot ship without it.

sshekhar563:
## Problem

`recover()` does `data = await provider.retrieve(source.reference); return data` with no check that `hash(data) == source_reference` for content-addressed tiers (IPFS/Arweave). The provider protocol's `verify()` only checks existence, and the CLI `recover` command writes the bytes straight to disk with no content validation.

A compromised gateway can return a different soul blob for a given CID and it is accepted. The whole point of content addressing is defeated without the check.

Closes #293.

## Root Cause

`EternalStorageManager.recover()` at `manager.py:107-109` trusts the provider to return authentic bytes. There is no content-address verification step between retrieval and return.

## Fix

1. **Extended `EternalStorageProvider` protocol** with:
   - `content_addressed: bool` — whether this tier uses content-addressing (CID = hash of data)
   - `compute_reference(data) -> str` — recompute the canonical reference from raw bytes

2. **Updated `recover()`** to verify retrieved bytes for content-addressed tiers:
   - After `provider.retrieve()`, if `provider.content_addressed` is True, compute the expected reference from the retrieved data
   - If it doesn't match `source.reference`, reject the data and try the next source
   - Non-content-addressed tiers (Arweave, Blockchain) skip the check

3. **Updated all providers**:
   - `MockIPFSProvider`: `content_addressed=True`, `compute_reference()` reuses `_generate_cid()`
   - `MockArweaveProvider`: `content_addressed=False` (tx IDs are sequential, not content hashes)
   - `MockBlockchainProvider`: `content_addressed=False`
   - `LocalStorageProvider`: `content_addressed=False` (reference includes soul_id, can't recompute from data alone)

## Tests Added

- `test_recover_detects_tampered_ipfs_content` — tampered IPFS data raises RuntimeError with "content-address mismatch"
- `test_recover_non_content_addressed_skips_check` — Arweave recovers without hash verification
- `test_content_addressed_property_on_providers` — all providers report correct `content_addressed` value
- `test_tampered_ipfs_falls_back_to_arweave` — tampered IPFS source is skipped, recovery succeeds via Arweave fallback

## Verification

- All 72 eternal storage tests pass (`tests/test_eternal/`)
- Ruff lint + format: All checks passed
- Backward compatible: `getattr(provider, "content_addressed", False)` safely handles legacy providers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,13 @@ Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - **Model Consolidation:** The dual `MemoryEntry` models in `runtime/types.py` and `spec/memory.py` have been consolidated into a single canonical source of truth in the spec layer. The runtime layer re-exports them for backward compatibility.
 - **Spec Documentation:** `SOUL-FORMAT-SPEC.md` section 6.4 updated to match the v0.4.0/v0.5.0 model additions (layer, domain, ingested_at, superseded, scope, user_id, etc.).
 
+### Eternal storage content-address verification (#293)
+
+- **Breaking:** `EternalStorageProvider` protocol now requires two new members: `content_addressed: bool` property (whether the tier uses content-addressing) and `compute_reference(data: bytes) -> str` method (recompute the canonical reference from raw bytes). Third-party providers written against the previous RFC-005 contract will fail `register()` until they add both members. Non-content-addressed tiers should return `False` and raise `NotImplementedError` from `compute_reference()`. See amended RFC-005.
+- **Security fix:** `EternalStorageManager.recover()` now verifies retrieved bytes against the content-addressed reference for tiers where `content_addressed is True`. A compromised gateway returning substituted data for a valid CID is detected and rejected. `register()` validates provider conformance at registration time (fail-closed).
+- **CLI fix:** `soul recover` now exits with code 1 when recovery fails (e.g. content-address mismatch). Previously it printed an error but exited 0, allowing `soul recover ... && deploy` to proceed after a detected substitution attack.
+
+
 ### Deprecated
 
 - **`soul remember` CLI alias** - deprecated in favor of `soul note`. Still supported in v0.5.x, and now explicitly scheduled for removal in v0.7.0.
--- a/rfc/RFC-005-ETERNAL-STORAGE-PROVIDER.md
+++ b/rfc/RFC-005-ETERNAL-STORAGE-PROVIDER.md
@@ -6,6 +6,8 @@
 **Status:** Draft -- Open for feedback
 **Author:** Soul Protocol Community
 **Date:** 2026-03-08
+**Amended:** 2026-08-15 (#293) — Added `content_addressed` property and
+`compute_reference()` method to prevent substitution attacks on `recover()`.
 
 ## Summary
 
@@ -44,6 +46,23 @@ class EternalStorageProvider(Protocol):
     @property
     def tier_name(self) -> str:
         """Name of this storage tier (e.g., 'ipfs', 'arweave')."""
+        ...
+
+    @property
+    def content_addressed(self) -> bool:
+        """Whether this tier uses content-addressing (e.g. IPFS CID).
+
+        When True, recover() verifies retrieved bytes by calling
+        compute_reference() and comparing to the stored reference.
+        """
+        ...
+
+    def compute_reference(self, data: bytes) -> str:
+        """Compute the canonical reference for data.
+
+        Only meaningful when content_addressed is True.
+        Non-content-addressed tiers must raise NotImplementedError.
+        """
         ...
 
     async def archive(self, soul_data: bytes, soul_id: str,
@@ -137,7 +156,9 @@ All non-local tiers use content addressing: the reference is derived from the da
 itself. This provides:
 
 - **Integrity verification.** Re-hash the retrieved data and compare to the reference.
-  If they match, the data is unmodified.
+  If they match, the data is unmodified. The `content_addressed` property and
+  `compute_reference()` method (added in #293) enable `recover()` to perform this
+  check automatically, rejecting substituted data from compromised gateways.
 - **Deduplication.** Archiving the same soul state twice produces the same reference.
   No wasted storage.
 - **Location independence.** The reference doesn't depend on where the data is stored.
--- a/src/soul_protocol/cli/main.py
+++ b/src/soul_protocol/cli/main.py
@@ -1050,6 +1050,7 @@ def recover(reference, tier, output):
             )
         except RuntimeError as exc:
             console.print(f"[red]Recovery failed:[/red] {exc}")
+            raise click.exceptions.ClickException(str(exc))
 
     asyncio.run(_recover())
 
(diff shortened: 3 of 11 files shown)


prakashUXtech (changes requested):
You avoided the trap I was most worried about here. Delegating to `provider.compute_reference(data)` instead of doing `sha256(data).hexdigest() == reference` in the manager is the right architecture — an IPFS CID isn't a bare sha256 (multihash + UnixFS DAG + CID version), and your design lets a real provider implement genuine CID computation without touching the manager. Putting the check in `recover()` rather than in each provider is also correct: one choke point, not N places to forget. And marking Arweave `content_addressed = False` is factually right, since a tx id is SHA-256 over the signed transaction, not the payload.

The blocker is that the gate is off by default.

```python
if getattr(provider, "content_addressed", False):
```

That third argument is a fail-open. A provider that simply *omits* the attribute skips verification silently — no error, no warning, bytes returned. I proved it with a provider implementing the protocol exactly as it existed before this PR, pointed at a hostile gateway:

```
recover() returned: b'ATTACKER-SUBSTITUTED-SOUL'
isinstance(provider, EternalStorageProvider): False
```

That's the vulnerability #293 asked to close, still fully open for the next provider someone writes. Note the second line: the `@runtime_checkable` protocol already *knows* the provider is non-conforming — `register()` just never asks. The PR body presents this as a feature ("safely handles legacy providers"), but backward-compatible fail-open is the anti-pattern for a security gate.

**Required:** make it mandatory. Validate in `register()` (the `isinstance` check already returns False correctly), or drop the `getattr` default so a missing attribute raises. At minimum a provider on a known content-addressed tier must hard-fail rather than skip. And add the test for it — a provider missing `content_addressed` must not silently skip verification. That test is what would have caught this.

Credit where due on the other two cases: an unrecognised reference or a raising `compute_reference` lands in the broad `except` and fails closed, and a genuine mismatch fails closed too. Those are right.

Two more worth addressing in this PR or an explicitly-filed follow-up:

**A detected tamper silently downgrades to an unverified tier.** On mismatch you `continue` to the next source, and `test_tampered_ipfs_falls_back_to_arweave` codifies it: IPFS tamper detected → falls through to Arweave → Arweave has no verification → those bytes return as success. The `errors` list is discarded once any source succeeds, so the caller learns neither that tampering was detected nor that what it got was never verified. An attacker who can substitute on the verified tier gets a free downgrade. At minimum this should log — there's no logging in the module at all, and a content-address mismatch is a security event.

**Arweave ends up with zero verification, and a test blesses it.** `test_recover_non_content_addressed_skips_check` asserts `recovered == b"different-payload"` — "a substituted Arweave payload is returned successfully" is now a named, passing test. Arweave transactions do carry `data_root`, a Merkle root over the data chunks, inside the signed tx, so authenticity *is* derivable — just not by a bare hash compare. Either rename that test to mark a known gap and file the follow-up, or don't close #293 on this PR.

Smaller: string equality isn't CID equality (CIDv0 vs v1, multibase, codec, chunker all change the string for identical bytes — it fails safe, but it'll false-reject the day a real gateway lands, so pin the canonical normalization in `compute_reference()`'s docstring); the CLI exits 0 on rejection so `soul recover … && deploy` proceeds after a detected tamper; `compute_reference()` returning `""` for non-content-addressed providers is a footgun (`raise NotImplementedError` is safer); the two new required protocol members are a breaking change to a publicly exported `@runtime_checkable` Protocol and want a CHANGELOG entry; and the file headers on all five changed files.

Fail-without-fix is solid — 3 of your 4 new tests genuinely fail on dev. Turn the gate on by default and this is done.

prakashUXtech (changes requested):
The blocker is properly closed, and you fixed it at the right layer. `register()` now rejects non-conforming providers outright, so the check can't be skipped by omission. I re-ran my probe from last round — the provider that previously returned `b'ATTACKER-SUBSTITUTED-SOUL'` is now rejected at registration, along with one that declares the flag but omits `compute_reference`, and one that sets the flag as an instance attribute rather than a class property. The `SECURITY:` warning log with a caplog test is a good addition too. And the tests aren't tautological: your test file run against dev's `manager.py` fails 6 of them.

Two things before merge.

**1. Enforcing conformance at `register()` is a breaking change to a published contract.** Any third-party provider written against RFC-005 as it currently reads will now fail to register — the two new required members (`content_addressed`, `compute_reference`) aren't in the RFC. That needs a CHANGELOG `Breaking` entry and an RFC-005 amendment in this PR; our rule is that a changed RFC contract gets amended alongside the code. This isn't a criticism of the fix, it's the cost of doing it properly.

**2. The CLI still exits 0 on a rejected recovery.** That was an explicit ask last round and is untouched — `soul recover ... && deploy` still proceeds after a detected substitution attack. One `sys.exit(1)` (or a `ClickException`).

Non-blocking, but worth resolving: you renamed the Arweave test to `..._known_gap`, which is honest, but it still asserts a substituted payload comes back successfully, and the docstring promises a Merkle `data_root` follow-up with no issue number. Please file that issue so the test points at something real — otherwise "known gap" is just a comment. Same for the tamper-downgrade path: a detected IPFS tamper still falls through to unverified Arweave bytes and returns them as success with `errors` discarded, so the caller can't tell. Your call whether that's acceptable, but it should be a decision rather than a default.

Headers on `spec/eternal/protocol.py` and the four provider files can ride along.

prakashUXtech (approved):
Both blockers are closed properly. The CHANGELOG entry does the hard part — it names both new protocol members, states plainly that third-party providers written against the old RFC-005 will fail `register()`, and tells them what to return for non-content-addressed tiers. RFC-005 is genuinely amended with the members rather than just a note. And `soul recover` now raises `ClickException`, so a detected substitution attack no longer exits 0.

That's the whole ask from last round, and combined with the earlier `register()`-time conformance guard this is a complete fix: the gate can't be skipped by omission, the break is declared, and the failure is scriptable.

Approving. Two loose ends worth tidying whenever, neither blocking:

The Arweave `data_root` gap still has no issue number. The test is honestly named `..._known_gap` and the docstring explains the Merkle-root reasoning, but there's no open issue to point at, so "known gap" currently means "a comment someone wrote". Please file it so the test references something real — that's half the value of naming it a known gap.

File headers on `spec/eternal/protocol.py` and the four provider files didn't get the `#293` line (`manager.py` did).

And a decision rather than a defect: a detected IPFS tamper still falls through to unverified Arweave bytes and returns them as success with `errors` discarded, so the caller can't tell it was downgraded. You've now made that explicit with the known-gap test, which is fine — I'd just want it to be a recorded decision rather than something we rediscover later.

PR #311, opened 2026-08-26
fix(#305): preserve created_at in memory_journal round-trip

### LINKED ISSUE #305: spike/memory_journal loses creation time after MemoryEntry consolidation
Follow-up from #300 (deferred because it's experimental `spike/` code, not the shipped runtime path).

After #285/#300 turned `MemoryEntry.timestamp` into a read-only property, `spike/memory_journal.py:417` still constructs `MemoryEntry(..., timestamp=datetime.fromisoformat(created_at), ...)`. Pydantic v2 silently ignores a kwarg that names a property, so the value is dropped and `created_at` falls back to `now()`. Every entry reconstructed from the journal store loses its stored creation time.

It stays green in CI because `recall()` orders by SQL `seq`, not the timestamp, and no test asserts timestamp fidelity.

**Fix:** `timestamp=` -> `created_at=` at spike/memory_journal.py:417 (one line). While there, `rfc/RFC-004-SOUL-FILE-FORMAT.md:170` still shows `MemoryEntry(content="Hello!")`, which now raises under the strict validator (same fix that landed in spec/container.py) — update it to include a `layer` or `type`.

sshekhar563:
Resolves #305.

## Problem

After #285/#300 turned `MemoryEntry.timestamp` into a read-only property, `spike/memory_journal.py:417` still constructs `MemoryEntry(..., timestamp=..., ...)`. Pydantic v2 silently ignores a kwarg that names a property, so `created_at` falls back to `now()`. Every entry reconstructed from the journal loses its stored creation time.

Additionally, `store()` never persisted the entry's `created_at` in the event payload — `_apply_remembered` used `event.ts` (always `now()`) as the SQLite `created_at`.

## Fix

1. `_row_to_entry` — `timestamp=` → `created_at=` (the actual Pydantic field)
2. `store` — persist `entry.created_at.isoformat()` in the event payload
3. `_apply_remembered` — use `p.get("created_at")` if available, fallback to `event.ts` for pre-fix events
4. `rfc/RFC-004-SOUL-FILE-FORMAT.md:170` — add `layer="episodic"` to `MemoryEntry` example (strict validator now requires `type` or `layer`)

## Testing

- Added `test_created_at_round_trip` — stores with past `created_at`, recalls, asserts timestamp survives
- All 24 existing journal tests pass
- Ruff check passed

### DIFF
--- a/rfc/RFC-004-SOUL-FILE-FORMAT.md
+++ b/rfc/RFC-004-SOUL-FILE-FORMAT.md
@@ -167,7 +167,7 @@ And the `SoulContainer` class provides a high-level API:
 
 ```python
 soul = SoulContainer.create("Aria", traits={"role": "assistant"})
-soul.memory.store("episodic", MemoryEntry(content="Hello!"))
+soul.memory.store("episodic", MemoryEntry(content="Hello!", layer="episodic"))
 soul.save("aria.soul")
 
 soul = SoulContainer.open("aria.soul")
--- a/src/soul_protocol/spike/memory_journal.py
+++ b/src/soul_protocol/spike/memory_journal.py
@@ -153,6 +153,7 @@ class JournalBackedMemoryStore:
             "emotion": entry.metadata.get("emotion"),
             "tags": entry.metadata.get("tags", []),
             "source": entry.source,
+            "created_at": entry.created_at.isoformat(),
         }
         event = EventEntry(
             id=uuid4(),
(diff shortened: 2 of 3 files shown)


prakashUXtech (approved):
Verified and merging. The new test genuinely guards the bug: with dev's `memory_journal.py` swapped in it fails with `created_at` coming back as now() instead of the stored 2025 value, and with the fix all 24 journal tests pass. Since `store()` and journal replay both route through `_apply_remembered`, persisting `created_at` in the payload fixes both paths in one place — nice.

One non-blocking gap if you ever touch this again: the `or event.ts` fallback for pre-fix events has no test, so a future refactor could drop it silently. Skipping the CHANGELOG is fine here since `spike/` is experimental.

